PR #33, opened 2018-12-22
Major cleanup of Krazo dependencies

chkal:
As we need to file some CQs for our dependencies, I did some cleanup to reduce the number of dependencies for the krazo-core module:

  * I replaced the umbrella `javaee-web-api` dependency with individual spec API JARs. A few of them were already released under the new `jakarta.*` groupId, so we don't need to file additional CQs for them. This will also allow us to update individual spec dependencies more easily.
  * There were some test scoped dependencies which were actually unused. I removed them from the pom.
  * We had a few Apache Deltaspike dependencies just for a single test which required to bootstrap a full CDI container. I updated the test to manually bootstrap Weld using the Weld API which is actually very easy. So I was able to get rid of the Deltaspike dependency.

I'll create CQs for the remaining dependencies later today and merge this pull request after these CQs are approved.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -175,39 +175,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.apache.deltaspike.core</groupId>
-            <artifactId>deltaspike-core-api</artifactId>
-            <version>1.6.0</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.deltaspike.core</groupId>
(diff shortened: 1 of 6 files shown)


chkal:
I also refactored the dependencies of the Jersey/CXF/RESTEasy modules. The additionak CQs:

  * [RESTEasy](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18596)
  * [CXF](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18597)

chkal:
All the required CQs were approved. So I'm merging this now!

Issue #21, opened 2018-12-07
CSRF token is only validated if there are @FormParams


mvcbot:
<a href="https://github.com/mvcbot"><img src="https://avatars2.githubusercontent.com/u/27966277?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mvcbot](https://github.com/mvcbot)**
_Friday Oct 23, 2015 at 05:54 GMT_
_Originally opened as https://github.com/mvc-spec/ozark/issues/79_

----

**Original issue OZARK-63 created by Christian Kaltepoth:**

The CSRF page token validation works fine in this case:

```java
@Controller
@Path("/foobar")
public class TweetController {

  @POST
  @CsrfValid
  public String post( @FormParam("text") String text ) {
    System.out.println("CSRF page token valid!");
    return ....;
  }

}
{code}

But if I remove the controller method parameter, the check seems to get skipped and the controller method is always executed. Even if the page token is missing.

{code:java}
@Controller
@Path("/foobar")
public class TweetController {

  @POST
  @CsrfValid
  public String post( /** empty **/ ) {
    System.out.println("CSRF page token valid!");
    return ....;
  }

}
```

mvcbot:
<a href="https://github.com/mvcbot"><img src="https://avatars2.githubusercontent.com/u/27966277?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mvcbot](https://github.com/mvcbot)**
_Friday Oct 23, 2015 at 12:43 GMT_

----

**Comment by Santiago Pericas-Geertsen:**

This is because the verification is done in a ReaderInterceptor. If there is no entity to read, the verification is skipped. This is a bit of an edge case in that regard, but perhaps we need to address it too.

gtudan:
Fixed in #43. CSRF-Validation is now done in a ContainerRequestFilter.

chkal:
Great! I also think that this shouldn't be an issue anymore. I updated the milestone to m05.

PR #39, opened 2019-01-26
#43: Fix CSRF-Validation on wildlfy

### LINKED ISSUE #38: CSRF token validation doesn't work on Wildfly
I didn't check this in detail, but the TCK tests for validating CSRF tokens in Wildfly are failing. It looks like submitting invalid tokens when CSRF protection is enabled doesn't result in a 401 status code as required by the spec. Perhaps the required validation isn't executed at all?

It would be great if somebody could have a look at this.

gtudan:
fixes #38


chkal:
Hey @gtudan,

thank you very much for the pull request. As mentioned before, I don't see any obvious problem with using a `ContainerRequestFilter` instead. I just checked the results of the TCK tests for this PR and the results are *interesting*:

  * The CSRF TCK tests are green for TomEE (CXF) now. That's awesome!
  * On Liberty (CXF), the tests results look much better now. Invalid tokens lead to a 401. That perfect. But some other CSRF tests are failing now. I guess this needs some further analysis.
  * On Wildfly missing tokens now also result in a 401. So the tests which were red before are now green. However, it looks like in case of successful CSRF token validation the submitted form parameters are not bound correctly, and we are getting null now. See the following test results for details:

```
[ERROR] Failures: 
[....]
[ERROR]   CsrfHeaderTest.submitValidTokenViaForm:97 
Expected: a string containing "Hi Alice!"
     but: was "<!DOCTYPE html>
<html lang="en">
    <body>
        <span id="message">Hi null!</span>
    </body>
</html>"
[ERROR]   CsrfVerifyDefaultConfigTest.submitFormWithAnnotationAndValidToken:60 
Expected: a string containing "Hi Alice!"
     but: was "<!DOCTYPE html>
<html lang="en">
    <body>
        <span id="message">Hi null!</span>
    </body>
</html>"
[ERROR]   CsrfVerifyExplicitConfigTest.submitFormWithAnnotationAndValidToken:60 
Expected: a string containing "Hi Alice!"
     but: was "<!DOCTYPE html>
<html lang="en">
    <body>
        <span id="message">Hi null!</span>
    </body>
</html>"
[ERROR]   CsrfVerifyImplicitConfigTest.submitFormWithAnnotationAndValidToken:60 
Expected: a string containing "Hi Alice!"
     but: was "<!DOCTYPE html>
<html lang="en">
    <body>
        <span id="message">Hi null!</span>
    </body>
</html>"
[ERROR]   CsrfVerifyImplicitConfigTest.submitFormWithoutAnnotationAndValidToken:89 
Expected: a string containing "Hi Charlie!"
     but: was "<!DOCTYPE html>
<html lang="en">
    <body>
        <span id="message">Hi null!</span>
    </body>
</html>"
[INFO] 
[ERROR] Tests run: 76, Failures: 7, Errors: 0, Skipped: 0
[INFO]
```

I wonder if this has something to do with the behavior I described [here](https://github.com/eclipse-ee4j/jaxrs-api/pull/695#issuecomment-440912357).

BTW: I created some first documentation about how to run the TCK tests here:

https://github.com/eclipse-ee4j/krazo/wiki/Running-the-TCK

chkal:
My current guess is that this isn't working correctly for RESTEasy:

https://github.com/eclipse-ee4j/krazo/blob/371ab7106ee13581c8ac0348184eb5c98e23e2cc/core/src/main/java/org/eclipse/krazo/security/CsrfValidateFilter.java#L104-L111

https://github.com/eclipse-ee4j/krazo/blob/371ab7106ee13581c8ac0348184eb5c98e23e2cc/core/src/main/java/org/eclipse/krazo/security/CsrfValidateFilter.java#L139-L141

Maybe RESTEasy uses the original input stream (the one we are consuming in the filter) for parsing the form to bind the `@FormParam` values and ignores the copy we store in the context in case of a non-ByteArrayInputStream?

gtudan:
Hmmm, to be honest I don't really get it. The input stream does get set to the context. 
My best guess is that resteasy relies on a ServletInputStream - maybe it uses the ReadListeners set there? I'll try to poke around and see if implementing this works.

I start understanding why the ReaderInterceptor was used here and what the problem is - while it is the intended place for accessing the entity, the reader only seems to get invoked if the message body is used in the controller (unmarshalling it to a java bean). If RestEasy notices that we don't care about the entity it will skip reading it and never invoke the listener.

chkal:
> Hmmm, to be honest I don't really get it. The input stream does get set to the context. My best guess is that resteasy relies on a ServletInputStream 

I really think it may be related to the behavior [I described here](https://github.com/eclipse-ee4j/jaxrs-api/pull/695#issuecomment-440912357). My guess is that RESTEasy parses `@FormParam` values from the same stream that is used to read entity parameters (and which is available from the context). 

The problem is if you consume the entity stream like we do it in the filter (or declare an entity parameter which triggers the consumption), RESTEasy will fail to read the stream again when it tries to get the values for `@FormParam` and therefore the values are `null`. IMO this is clearly an RESTEasy bug.

> I start understanding why the ReaderInterceptor was used here and what the problem is - while it is the intended place for accessing the entity, the reader only seems to get invoked if the message body is used in the controller (unmarshalling it to a java bean).

I think `ReaderInterceptor` was used as it was meant to be an API for modifying/reading the request entity body. Back then Ozark only supported Jersey and it was working fine there. But only because of the specific handling of Jersey regarding reader interceptors.

I really think that using an `ReaderInterceptor` won't work very well for us, because not all implementations execute the interceptor if there is no entity parameter (again, see the discussion [here](https://github.com/eclipse-ee4j/jaxrs-api/pull/695)) I actually like the idea of using a request filter instead. The fact that this doesn't work for RESTEasy could really be just a bug on their side.

chkal:
This is really old, but sounds exactly like this issue:

https://issues.jboss.org/browse/RESTEASY-567

chkal:
Hey @gtudan,

first of all: Thanks a lot for working on this. I'm happy to see that you are spending some time to dig into this issue which is a major blocker at the moment.

Unfortunately I've some bad news. It looks like the TCK is failing now on Glassfish with similar errors:

```
[ERROR]   CsrfVerifyImplicitConfigTest.submitFormWithoutAnnotationAndValidToken:89 
Expected: a string containing "Hi Charlie!"
     but: was "<!DOCTYPE html>
<html lang="en">
    <body>
        <span id="message">Hi null!</span>
    </body>
</html>"
```

To be honest, I think your first approach (replacing the reader interceptor with a container request filter and parsing the form manually from the input stream provided from the context) is the better way to implement this feature. 

You are correct, that `HttpServletRequest.getParameterMap()` allows us to access the submitted form parameters, but unfortunately calling this method also forces the servlet container to read the full request body (and therefore the entity stream) which breaks all subsequent attempts to read it (like Jersey seems to try it). 

The advantage of manually parsing the `InputStream` provided by `ContainerRequestContext` is that it also allows us to copy the input stream and pass the copy to `setEntityStream()` so the stream can be consumed again by the JAX-RS runtime later on without issues. In theory this is the best way to go. The only problem is that RESTEasy doesn't handle this case correctly. However, I think this is a clearly a bug that needs to be reported. If RESTEasy cannot fix this, we could think about introducing a SPI which allows `krazo-resteasy` and `krazo-jersey` to implement the access to the form parameter in different ways (RESTEasy by using the `HttpServletRequest`, Jersey by parsing it manually and replacing the entity stream in the context).

It looks like RESTEasy has [an Arquillian-based test suite](https://github.com/resteasy/Resteasy/tree/master/testsuite/integration-tests) which should allow creating failing test very easily. In theory, we just have to create a test which replaces the InputStream in the `ContainerRequestContext` and check whether the changed data is reflected in the values injected via `@FormatParam`.

What do you think?

gtudan:
Murphy strikes again :-D
Well, I guess we're out of luck then. I'll try to reopen the Resteasy issue and see what happens. 

Creating an SPI for this sounds like a good idea - there are quiet a lot of older JBoss and wildflys out there and I would hate leaving them behind. I'll see what I can do - maybe use the original implementation as default and drop an alternative in the Resteasy package? We can just delete the alternative implementation once RestEasy gets this fixed.

gtudan:
+1 - I'll look into it

chkal:
@gtudan 

Thanks a lot! This looks great! I just checked the TCK tests and all CSRF tests now pass on Glassfish, Wildfly and TomEE. For some reason the tests for checking CSRF tokens via headers fail on Liberty, but I'm not sure if this is related to this PR at all. However, we can fix this later. Having green tests on the other containers is awesome and a huge step forward.

Regarding the license headers. Most files are fine, with a few exceptions.

  * Some new files seem to use the old header. Could you update them to the new "Eclipse Krazo committers and contributors" header. Maybe they are not all completely new, because they have been moved and/or renamed, but I think in this case we could update the header.
    * DefaultFormEntityProvider.java
    * DefaultFormEntityProviderTest.java
    * CsrfIT.java
  * Could you update 2018 to 2019 in this file:
    * WebArchiveBuilder.java

Sorry, I know that keeping copyright headers up to date is a pain. I guess we still need to find a good workflow here. IntelliJ seems to have some basic support for this. Or maybe a Maven plugin that can do this.

I would love to merge this as soon as possible. Thanks a lot for your help. :beers:

gtudan:
Cool - I updated the copyright headers.

chkal:
Awesome! I guess this is ready to be merged now! Thanks a lot!

BTW: I guess this will also fix #21.

PR #40, opened 2019-01-31
Major refactoring to support firing events as required by the spec

chkal:
Hey all,

this PR contains a few refactorings which allows us to fire `AfterControllerEvent` and `AfterProcessViewEvent` even if the corresponding controller/view fails. This is required by the spec and a few TCK tests are currently failing because of this. Here is what I did:

  * There are now two `ContainerRequestFilters` (one pre-matching and a post-matching one) which use JAX-RS's `@Context` to get a few JAX-RS managed objects (like `UriInfo`, `ResourceInfo`). These objects are then passed to a CDI-managed producer which allow us to inject those JAX-RS objects via CDI.
  * I added a new CDI interceptor called `AroundControllerInterceptor`. The interceptor is added automatically to all controller methods and therefore allows us to execute code before and after the controller invocation. This is much more reliable, then using the JAX-RS filter SPI, especially for firing events.
  * There is a new class called `RequestLifecyle` which I think could become the central class for Krazo to implement the various steps required as part of the JAX-RS request lifecycle. JAX-RS providers and CDI interceptors would just call methods on this class which implement the desired behavior. I'm planning to move more code to this architecture in the next days. The advantage is that we get a much clearer view of the various steps that need to be performed, and we can ensure correct ordering. It will also reduce the overall number of JAX-RS providers we need to implement, because all coordination would be done by the `RequestLifecycle` class. This is somehow similar how JSF implements the request lifecycle.
  * There is now a new class `EventDispatcher` which is used to fire most of the events defined by the spec. This way the event handling isn't spread over different JAX-RS providers anymore.

Two more notes:
  * I added a few new archive providers which build the Krazo archives from `target/classes` which simplifies running the TCK while working on the Krazo code.
  * I modified the Travis config  to allow failures for the `glassfish-module` build for now. This build is failing with some weird errors at the moment. I'll try to fix this later, but I guess we should update to Eclipse Glassfish 5.1.0.Final before working on this.

Let me know if you have any questions.


core/src/main/java/org/eclipse/krazo/cdi/AroundController.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright © 2017, 2018 Ivar Grimstad
```
ivargrimstad: I think this should be ...2017, 2019... according to mail discussions on the Jakarta EE Community mailing list...

core/src/main/java/org/eclipse/krazo/cdi/AroundControllerInterceptor.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright © 2017, 2018 Ivar Grimstad
```
ivargrimstad: Same as above

core/src/main/java/org/eclipse/krazo/cdi/AroundController.java
ivargrimstad: ...or actually only ...2019..., I guess since it being a new file... the changed files should be ...2017, 2019...

core/src/main/java/org/eclipse/krazo/lifecycle/EventDispatcher.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Inject
+    private Event<MvcEvent> mvcEventDispatcher;
+
+    void fireBeforeControllerEvent() {
+        if (KrazoCdiExtension.isEventObserved(BeforeControllerEvent.class)) {
+            final BeforeControllerEventImpl event = new BeforeControllerEventImpl();
+            event.setUriInfo(uriInfo);
+            event.setResourceInfo(resourceInfo);
+            mvcEventDispatcher.fire(event);
```
ivargrimstad: Since we now are on CDI 2....wouldn't it make sense to use fireAsync(.. ?

tck/src/main/java/org/eclipse/krazo/tck/tomee/TomeeDevArchiveProvider.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright © 2017, 2018 Ivar Grimstad
```
ivargrimstad: 2019

tck/src/main/java/org/eclipse/krazo/tck/wildfly/WildflyDevArchiveProvider.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright © 2017, 2018 Ivar Grimstad
```
ivargrimstad: 2019

tck/src/main/java/org/eclipse/krazo/tck/liberty/LibertyDevArchiveProvider.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright © 2017, 2018 Ivar Grimstad
```
ivargrimstad: 2019

tck/src/main/java/org/eclipse/krazo/tck/glassfish/GlassfishDevArchiveProvider.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright © 2017, 2018 Ivar Grimstad
```
ivargrimstad: 2019

core/src/main/java/org/eclipse/krazo/lifecycle/RequestLifecycle.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright © 2017, 2018 Ivar Grimstad
```
ivargrimstad: 2019

core/src/main/java/org/eclipse/krazo/lifecycle/EventDispatcher.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright © 2017, 2018 Ivar Grimstad
```
ivargrimstad: 2019

chkal:
Hey @ivargrimstad! Yes, you are absolutely correct, I forgot to update the year in the copy right headers.

But actually I'm not even sure if the Maven plugin we are currently using allowed to have different headers per file. AFAIK it just compares against the template file (which you can find [here](https://github.com/eclipse-ee4j/krazo/blob/master/etc/license-header.txt)). I've to check that.

core/src/main/java/org/eclipse/krazo/jaxrs/PostMatchingRequestFilter.java
```diff
@@ -0,0 +1,82 @@
...
+        JaxRsContextProducer contextProducer = CdiUtils.getApplicationBean(JaxRsContextProducer.class)
+                .orElseThrow(() -> new IllegalStateException("Cannot find CDI managed JaxRsContextProducer"));
+
+        // store JAX-RS context objects so we can produce them via CDI
+        contextProducer.setConfiguration(configuration);
+        contextProducer.setRequest(request);
+        contextProducer.setResponse(response);
+        contextProducer.setApplication(application);
+        contextProducer.setUriInfo(uriInfo);
+        contextProducer.setResourceInfo(resourceInfo);
```
gtudan: Do we really need to set all these objects again, or wouldn't it be enough to set `resourceInfo`?
Does the context change during matching?

chkal: Good point! 

Actually the only object we don't get in the pre-matching filter and for which we need the post-matching filter is the `ResourceInfo` instance. So you are correct, that all other attributes have already been set in the pre-matching filter and setting it again in the post-matching filter is redundant.

Actually I did this because of the problems with the `glassfish-module` tests. In the test failures it looked like the pre-matching filter wasn't executed and therefore `UriInfo` was never stored in the producer. My first guess was that the pre-matching filter isn't executed, and so this was my try to fix it. But it looks like both filters aren't executed in the `glassfish-module` build. I've no idea why and will investigate some more time later.

However, I think that doing it twice does no harm!?

chkal:
Hi everyone! I just pushed an updated version of this PR containing the following changes:

 * I updated the copyright year in all modified files. For new files I added a more generic copyright notice which describes the copyright holder as "Eclipse Krazo committers and contributors". I think this should be the default license header for all new files.
 * As we now have different copyright headers in our files, I removed the old license Maven plugin. This plugin wasn't very flexible, and we should look for an alternative soon.

core/src/main/java/org/eclipse/krazo/jaxrs/PostMatchingRequestFilter.java
gtudan: I was just wondering. You‘re right, this diesn‘t hurt and if it makes testing easier it’s fine.

chkal: I guess we should clean this up (remove the duplication) as soon as we have more green builds, and can verify, that it works across all environments.

Issue #41, opened 2019-02-01
Support PUT / PATCH / DELETE methods in form


erdlet:
To be able to follow the REST principles in an MVC application it would be great to have a kind of interceptor which converts POST requests to e.g. PUT by some hidden form field's value.

I thought about something like this (currently some pseudo-code, I'm working on a JAX-RS based solution as my first approach runs only on a simple Filter):

```
@PreMatching
public class FormMethodFilter implements ContainerRequestFilter {

  @Override
  public void filter(ContainerRequestContext requestContext) throws IOException {
    if (isFormData(requestContext)) {

      // use EntityStream to receive form data
      String formData = getFormDataFromRequest(requestContext)

      // get hidden field value (or null if not set)
      String hiddenMethod = getHiddenMethod(formData)
     
     // replace current method with hidden one
     if (hiddenMethod != null) {
        requestContext.setMethod(hiddenMethod)
      }
    }
  }
}
```

erdlet:
I added my current implementation to my fork. Unfortunately, when I access the EntityStream of the `ContainerRequestContext`, any form parameter value is null in the controller. When I test this filter with a hard-coded method (and skipping the EntityStream processing) anything works. Seems there is some behavior regarding this EntityStream which I'm not aware of.

chkal:
Hey @erdlet,

reading from the entity stream is problematic in some cases, because it basically prevents other components to read it again. We have similar issues with the CSRF protection, which also needs to read the form (by consuming the entity stream) which breaks `@FormParam` in Wildfly.

You could have a look at #43 for details. @gtudan is currently working on an SPI which should fix these issues.

erdlet:
Hey @chkal,

thanks for the hint. I just had a look on the PR for #43 and I guess it doesn't make sense to
continue on the JAX-RS based form filter until the CSRF issue is solved. 

I'll watch the #43 progress, so I can implement the form method filter with the SPI when it's finished.

chkal:
@erdlet Ok, great! :+1:

chkal:
Fixed via #45

PR #45, opened 2019-02-07
[#41] Add support for PUT PATCH and DELETE method in forms

erdlet:


erdlet:
Just checked the build and I found out, that the failing test for the `DefaultFormEntityProvider` can fail in case it contains the byte representation of `=`. Opened issue #46 because it doesn't belong directly to this issue.

Daniel-Dos:
great @erdlet . 

thanks .

erdlet:
@Daniel-Dos 

you're welcome :)

core/src/main/java/org/eclipse/krazo/forms/HiddenMethodFilter.java
```diff
@@ -0,0 +1,101 @@
...
+    return MediaType.APPLICATION_FORM_URLENCODED_TYPE
+        .isCompatible(requestContext.getMediaType());
+  }
+
+  private String getHiddenMethod(final Form form) {
+    String hiddenMethod = null;
+    final List<String> hiddenFieldValues = form.asMap().getOrDefault(HIDDEN_METHOD_NAME, Collections
+        .emptyList());
+
+    if (!hiddenFieldValues.isEmpty()) {
```
ivargrimstad: Should there be some check for which methods that are allowed (i.e. check for PUT, DELETE, PATCH)?

erdlet: I didn't add this check intentionally, as non-HTTP verbs should fail automatically by some exception (tried it yesterday got an OperationNotSupportedException). Sure, OPTION and HEAD don't really make sense as form method, but I think that the developer using this feature should be aware of that.

ivargrimstad: Makes sense. Thanks!

Daniel-Dos:
I think a good option to add something on the website to RI resources : 

https://www.mvc-spec.org/krazo/ 
https://github.com/mvc-spec/www.mvc-spec.org

chkal:
@gtudan I agree that the mvc-spec.org site would be a great place to provide more details about such Krazo-specific features. Maybe we need some kind of basic structure for a Krazo documentation first?

chkal:
I'll merge this now. Thanks @erdlet for implementing this feature! :beers:

PR #49, opened 2019-02-12
Removed Groovy extension

chkal:
The Eclipse IP team detected a problem regarding our dependency on Groovy 2.5.3. It looks like Groovy depends on ANTLR2 which is not approved for distribution by the Eclipse Foundation. 

Please refer to [this CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18836) for details.

Therefore, this PR removes the Groovy Extension module. The code is still available in the old Ozark repository. So if anyone really wants this extension, we could move the code to a separate GitHub repo outside of the Eclipse Foundation.

If anybody thinks that I filed the CQ not correctly, feel free to speak up. My understanding is that we only depend on the Groovy Templates Maven dependency, but we also import classes from Groovy Core, which introduces the licensing issue described above.


chkal:
To be honest, I'm a bit confused right now. Looks like I didn't understand the request by the IP team in the CQ correctly. The [latest comment](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18836#c13) sounds like we **may** even be fine with the Groovy dependency, as long as we don't use any ANTLR code directly!?!?

Anyway, I already merged this PR. Some maybe we should leave it as it is for now!?!

If anyone really wants to keep the Groovy Template engine, feel free to open a new CQ and try to get it approved. I still don't fully understand what exactly we need to get an approval for: Groovy Templates, Groovy Core, ANTLR, or even more?

PR #59, opened 2019-04-11
Patch Jersey version in Glassfish 5.1.0 for TCK

chkal:
I adjusted our test script to replace one of the Jersey modules shipping with Eclipse Glassfish 5.1.0 with a patched version. Let see if this fixes some of the errors we get when running the TCK against Eclipse Glassfish 5.1.0.

Travis, please test! :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,10 @@ jobs:
       before_script: .travis/docker-payara.sh
       script: mvn -P${TYPE} --projects testsuite clean verify
     - stage: test
-      env: TYPE=tck-glassfish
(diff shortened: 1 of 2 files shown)


chkal:
Sorry, I forgot to mention what the fix is about. The patched JAR file was built from Jersey 2.28 + the fix for eclipse-ee4j/jersey#3670.

My first impression is that the build is much more reliable now. We have two test failures which are most likely other problem. I'll rerun the build a few times to make sure that the random failures described in the issue above are gone now.

chkal:
I can confirm that we have more green tests. So I'll merge this now.

PR #61, opened 2019-04-12
Ensure ViewResponseFilter is executed only once

chkal:
For some reason Jersey 2.28 executes our view filter twice. This results in duplicated MvcEvents being fired, which results in TCK failures. This workaround fixes this issue.


gtudan:
Quiet a hack - but I guess it works and get's the job done 😄 🤷‍♂️

chkal:
@gtudan I know. This is a bit dirty. :-)

But to be honest, I don't want to spend many many days debugging the Jersey code which seems to be the root cause of the issue. The last time I reported such an issue to the Jersey team, it took > 14 month to get a fix. :cry: 

That's why I think that such a workaround is acceptable for now. :-)

Issue #28, opened 2018-12-07
Migrate examples to test suite


mvcbot:
<a href="https://github.com/gtudan"><img src="https://avatars1.githubusercontent.com/u/419425?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [gtudan](https://github.com/gtudan)**
_Saturday Sep 29, 2018 at 15:39 GMT_
_Originally opened as https://github.com/mvc-spec/ozark/issues/209_

----

It would be great to have am integration testsuite that runs locally, so that we can make sure we don't break anything. Currently the tests are more examples than tests as building, deploying and testing them is quiet difficult.

See here for the discussion on the mailing list: https://groups.google.com/d/msg/jsr371-users/NZXgRX9610E/jb1l706fCQAJ

mvcbot:
<a href="https://github.com/chkal"><img src="https://avatars2.githubusercontent.com/u/159245?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [chkal](https://github.com/chkal)**
_Sunday Sep 30, 2018 at 07:09 GMT_

----

A big +1 for this! ;-)

One thing I would like to mention. As you may know we are currently in the process of migrating Ozark to the Eclipse Foundation (see the [Eclipse Krazo proposal](https://projects.eclipse.org/proposals/eclipse-krazo)). As part of this migration, we have to check the licenses of our dependencies to make sure that all the license are allowed (see #179). It MAY be possible, that we need to drop some extensions because they have dependencies with disallowed license. But that's not official yet. But I guess if we start to migrate the examples to the new test suite module, we should prefer to migrate example apps which test core features and example apps which test ASL2-licensed extensions first. Just to make sure we don't spend too much time on migrating apps which we have to remove as part of the migration anyway.

gtudan:
We've moved quiet a few examples to the test suite by now, focusing on stuff that were extensions or implementational details of Krazo and hence were not covered by the TCK. 

I think most of what is left is already covered by tests from the TCK. So I would like to ask the other developers what we should do with them. Should we:

1. still port them to the test suite? Maybe as starting point for more in-depth tests?
2. leave them as-is to have some examples that could serve as starting points for the users?
3. Remove them, since they served their purpose as early test-suite and are superseded by the TCK?

chkal:
@gtudan Thanks for bringing this up!

[...]

PR #68, opened 2019-04-28
Migrate all the extension examples to the testsuite

gtudan:
This migrates all the examples for extensions to the testsuite from #28 

There are still some issues on containers that I would like to fix before merging this. At least the wildfly issues look like quick fixes (mostly classloader issues).

I found a dependency to groovy-all in the jsr223-example. Do I remember right that this had an incompatible license? I removed it for now, as it's only a small test-case.


chkal:
> This migrates all the examples for extensions to the testsuite from #28

That's awesome! Thanks! :beers: 



> I found a dependency to groovy-all in the jsr223-example. Do I remember right that this had an incompatible license? I removed it for now, as it's only a small test-case.

Oh, looks like I missed this dependency. Feel free to remove it in this case.

gtudan:
Okay, the test-suite passes on Payara.
Wildfly needs more work (classloading is a bit difficult here). I'll file an issue for it.

chkal:
@gtudan Do you think we should merge this one now and address the Wildfly issues later or whould we first try to fix the Wildfly issues as part of this pull request?

chkal:
Of course I would prefer the latter. But if this isn't easy, we could also merge the PR first and address the Wildfly issues later.

gtudan:
Hi Christian! I'm afraid fixing the failing tests on wildly will take a while, as there are some nasty hacks required. I.e. the Asciidoctor-Example requires a custom module to be preinstalled, which quiet hard to do with docker and arquillian. https://github.com/asciidoctor/asciidoctorj/blob/26641dbe1510a83fbb9c5cab3974a8736b385bdd/README.adoc#running-asciidoctorj-on-wildfly-as

I'd prefer to merge this now to make sure we don't break anything before the release and open an issue for the failing tests on WF.

gtudan:
The issue for the failing tests is #70

chkal:
@gtudan Thanks a lot! In this case I'm fine with merging soon.

But one issue came to my mind. I guess you added a few new dependencies for the new test suite, correct (Arquillian adapters or something similar)? In this case we have to file a few CQs before integrating the 3rd party dependencies. I can help with this. I just need some time to have a closer look at the pom diffs.

gtudan:
Actually no, we layed the foundations for the test suite before we moved to eclipse, so everything was in place when the CQs were filed. This change does not add any dependencies that were not already in the example projects. But of course I‘ll double check the CQs to see if everything has been approved. I‘d highly appreciate your help if we find anything missing!

gtudan:
Looks like we’re lucky, all of the dependencies for the testsuite were in the test CQ:

https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18898

chkal:
@gtudan Awesome! Thanks for checking. In this case I'm fine with merging now. Thanks a lot for working on this.

PR #69, opened 2019-05-05
Added fix for RESTEASY-2222 to patched wildfly profile

chkal:
I added another quickfix to the `tck-wildfly16-patched`. And: Good news! The TCK passes now! :tada: 

The fix is a simple workaround for [RESTEASY-2222](https://issues.jboss.org/browse/RESTEASY-2222). Let's hope they are willing to accept a corresponding patch.

@gtudan IIRC we talked about some Wildfly specific CSRF failure at JavaLand? Do you remember any details about this failure? For some reason the TCK is green now!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,6 @@ jobs:
   allow_failures:
     - env: TYPE=tck-glassfish51-vanilla
     - env: TYPE=tck-wildfly16-vanilla
-    - env: TYPE=tck-wildfly16-patched
     - env: TYPE=tck-tomee
     - env: TYPE=tck-liberty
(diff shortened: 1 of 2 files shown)


gtudan:
mhhh... I think we were talking about CXF back then, which still has quiet a few quirks around CSRF. Wildfly looked good after we fixed that issue with FormParams and the CSRF-Filter.

chkal:
> mhhh... I think we were talking about CXF back then, which still has quiet a few quirks around CSRF. Wildfly looked good after we fixed that issue with FormParams and the CSRF-Filter.

Ah, ok. I thought it was about Wildfly. :laughing:

PR #75, opened 2019-06-21
Run testsuite against TomEE

gtudan:
Still working on improving the testsuite and making sure our extensions work with different providers.
This runs the integration-tests from the testuite against TomEE 7, so we get some coverage for CXF.

There are failures, but only two. I'm a pleasantly surprised :-)

* The Pebble Ext doesn't work - I will try to look into it.
* The hidden forms IT fails when using the patch-method. I'll file an issue for it. @erdlet could you please take a look?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,10 @@ jobs:
       before_script: .travis/docker-payara.sh
       script: mvn -P${TYPE} --projects testsuite clean verify
     - stage: test
+      env: TYPE=testsuite-tomee
+      before_script: .travis/docker-tomee.sh
+      script: mvn -P${TYPE} --projects testsuite clean verify
+    - stage: test
       env: TYPE=tck-glassfish51-vanilla
       script: .travis/tests.sh ${TYPE}
     - stage: test
@@ -43,4 +47,5 @@ jobs:
     - env: TYPE=tck-tomee
     - env: TYPE=tck-liberty
     - env: TYPE=testsuite-wildfly
+    - env: TYPE=testsuite-tomee
     - env: TYPE=glassfish-module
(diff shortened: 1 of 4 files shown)


erdlet:
Hey @gtudan,

I'll have a look on this test. The filed issue is #76 I guess?

gtudan:
Hi Tobias,

Right, that’s the one. Thanks for looking into it!

- Gregor

gtudan:
Okay, will do.
@chkal: could you please update the CQ for the Arquillian Adapter to the newest Version (8.0.0.M3)?

chkal:
Ok, sure. Although I think I need to create a separate one. Will do this
later today.

Gregor Tudan <notifications@github.com> schrieb am Sa., 29. Juni 2019,
13:19:

chkal:
I just created [CQ20315](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20315) and closed the old one.

chkal:
Sorry, I forgot to mention that I filed the CQ for version 8.0.0.M3, which is the latest one available.

chkal:
The new CQ is already approved. Thanks @ivargrimstad. :smile:

gtudan:
Okay, TomEE is updated to 8 and the tests mentioned above work fine.

BUT: since the docker image for 8 uses Java 11 the Jetbrick test now fails - it's a known issue from the wildfly test suite (which also uses Java 11). I'll try to get it fixed.

chkal:
> Okay, TomEE is updated to 8 and the tests mentioned above work fine.

:+1: 

> BUT: since the docker image for 8 uses Java 11 the Jetbrick test now fails - it's a known issue from the wildfly test suite (which also uses Java 11). I'll try to get it fixed.

Too bad. Thanks for looking into this.

chkal:
#82 needs to be rebased, so we get a green build and can merge it. After that, we need to rebase this one against master to get a green build. Well, that's complicated! :laughing:

Issue #76, opened 2019-06-21
Testsuite failures on TomEE


gtudan:
There are testsuite failures on TomEE:

- [x] The pebble IT fails to deploy.
- [x] The HiddenFieldsIT gets a HTTP 405 when the form is submitted using patch
- [x] The Jetbricks IT fails with TomEE 8 because that image uses Java 11

erdlet:
Hi all,

I debugged the error and I think this is a problem of TomEE itself.

TomEE 7.1.0 is using `cxf-rt-frontend-jaxrs:3.1.15` for JAX-RS, which depends on JAX-RS 2.0 (https://mvnrepository.com/artifact/org.apache.cxf/cxf-rt-frontend-jaxrs/3.1.15).

Unfortunately, the PATCH method is only supported since JAX-RS 2.1. This is the reason, why it works on Wildfly and Jersey, as both are already based on JAX-RS 2.1. I also tested against the upcoming TomEE 8 which uses the above mentioned dependency in version 3.3.1 (with JAX-RS 2.1) and there it works like a charm.

So I think there might be two ways to proceed:

1) Do nothing and don't support this old CXF version (we use in our dependencies CXF 3.3.1 everywhere) -> wait for TomEE 8
2) Don't support PATCH method in the HiddenMethodFilter and generate a nicer error (I really dislike this as it limits Krazo on newer servers)

What do you think?

gtudan:
Hi Tobias,

Thanks for looking into this and good find on JAX-RS 2.0.

I think it makes sense to disable this test case for TomEE 7 as both CXF and Krazo work as designed. So +1 for Option 1 😊

chkal:
I think we should go with option (1) for the following reasons:

  * MVC 1.0 requires JAX-RS 2.1
  * I already reported a few CXF issues which were (IIRC) fixed in recent releases. So I'm pretty sure that users would run into various other issues if they are using older CXF versions.

BTW: Currently we are [running the TCK](https://github.com/eclipse-ee4j/krazo/blob/d944a8cb1bf412da8a38e603cc95bf165ccb8b18/tck/pom.xml#L197) against TomEE 8.0.0-M2.

gtudan:
Testsuite is merged and green

PR #77, opened 2019-06-21
Make the pebble view engine proxyable to fix a test failure on TomEE

gtudan:
Fixes one off the test failures from #76


chkal:
LGTM

PR #83, opened 2019-07-04
fix TLS-Errors in the payara testsuite

gtudan:
write the payara certificate for the testsuite into the default keyst…ore instead of a seperate one. This prevents other SSL connections to validate certificates


chkal:
I'm wondering if we should move the `keytool` command to `tests.sh` instead. Doesn't this make more sense? Modifying the JDK default key store feels wrong. :-)

Anyway, I'll merge this now, so we get green builds again. We can think about how to handle this in the future later...

PR #84, opened 2019-07-06
Removed tck-glassfish51-vanilla profile

chkal:
Glassfish 5.1 is known to not work correctly with Krazo because of known issues in Jersey 2.28 which have been fixed in 2.29. So I don't think it makes sense to keep the `tck-glassfish51-vanilla` anymore. Hopefully there will be a new Glassfish release soon with an updated Jersey version.


chkal:
Looks like the build is failing because the Glassfish download fails at the moment... I guess the EF has some technical problems at the moment.

PR #87, opened 2019-07-26
34 custom csrf mappers

erdlet:
At the moment, Krazo assumes that every response entity
has to be a view which can be processed by a ViewEngine.
In case there is e.g. some custom exception mapper who wants to
return a plain text message, the Request stops with an exception.

To avoid this, Krazo can now be configured to either use this strict
evaluation or some less strict which enables Krazo to return entities
which are not views. To enable this, the config
'org.eclipse.krazo.strictViewResolution' has to be set to false.


chkal:
Thanks a lot! I'll try to find some time to review the PR in detail ASAP.

But maybe we have to wait a bit before merging it. As Ivar [mentioned on the mailing list](https://www.eclipse.org/lists/krazo-dev/msg00058.html), we are currently working on fixing some copyright header and license issues to comply with the EF requirements and therefore need to bulk-update a lot of files.

erdlet:
Sure that's ok :+1:

core/src/main/java/org/eclipse/krazo/core/ViewResponseFilter.java
```diff
@@ -198,17 +217,33 @@ static String appendExtensionIfRequired(String viewName, String defaultExtension
...
+    private void fillResponseEntity(final String view, final ContainerResponseContext responseContext) {
+        if (krazoConfig.isStrictViewResolution() || isViewTemplate(view) || isRedirect(view)) {
+            responseContext.setEntity(new Viewable(view), null, responseContext.getMediaType());
+        } else {
+            responseContext.setEntity(view, null, responseContext.getMediaType());
+        }
+    }
+
+    static boolean isViewTemplate(final String view) {
+        return VIEW_TEMPLATE_PATTERN.matcher(view)
```
gtudan: I‘m not quiet sure about this. 
At the moment views are not required to have an extension (the view response filter will add the configured default extension). Maybe we could use the Content-Type from the response instead:

If the flag is set and Content-Type is TEXT/PLAIN then don‘t run the entity through a view engine?

erdlet: OK, I wasn't aware of this default extension mechanism. Thanks for the hint :+1: 

I agree with the idea, but I'm not sure if a filter on `TEXT/PLAIN` is enough, as some people may want to return a completely different content type (I don't know why someone would do that, but we all know how software development works ;) ). Maybe we shall only process the template if we have `TEXT/HTML` as content type, which is set as default if nothing different is declared on the controller method, to ensure we really process a view.

What do you think about this?

gtudan: Mhh... that really is a tough one 🤔 

The most forward approach to this would probably be to wrap the entity in some kind of value class (`NoViewResponse`), but I don't see how we can get this done without adding this class to the API.

Maybe we could handle this through a special inline-template engine and string-formats?
So one could return a string with a special prefix (`"inline: This is an exception`) and the new engine just strips the prefix and returns the rest? 

The engines usually implement their `supports` method by the template extension, but I don't see why this should not work with a prefix.

erdlet: I think that should work. Would be the easiest way until we are able to extend the API with some class like the mentioned `NoViewResponse`. I'll try it tomorrow afternoon :+1:

gtudan: Sounds good!
But please don't get me wrong: I was just thinking out loud. Maybe the idea isn't that good and Christian will hate me for proposing API extensions 😉

chkal: I don't think that adding new class to the API is really an option, because we cannot introduce it as part of MVC 1.0 anymore.

And my current hope is, that we can address multiple of the current issues regarding exception handling with a single new concept. Maybe #94 is a good start.

I don't have much time to write down my thoughts now, but I'll try to do so later today. Stay tuned.

erdlet: So I tried the idea of an engine which checks for the `inline:`prefix. All tests are working when I use this solution (I'm not sure, if the override of the response's body is done correct here):

```
@Priority(ViewEngine.PRIORITY_BUILTIN)
public class InlineValueViewEngine implements ViewEngine {

    @Override
    public boolean supports(String view) {
        return view.startsWith("inline:");
    }

    @Override
    public void processView(ViewEngineContext context) throws ViewEngineException {
        final HttpServletResponse response = context.getResponse(HttpServletResponse.class);

        try {
            response.getWriter().write(context.getView().split(":")[1]);
            response.getWriter().flush();
        } catch (final IOException e) {
            throw new ViewEngineException("could not process inline view", e);
        }
    }

}
```
With this `ViewEngine` we could remove the additional `KrazoConfig` entry and it is more transparent what is happening than before.

gtudan: I like it - looks like a useful feature. This would also work on controllers for rendering "raw" output. 👍 

I'd prefer this kind of implementation as it doesn't add more complexity to the rendering pipeline. It's also pretty lightweight in terms of performance.

@chkal what do you think?

chkal: Hey all. As mentioned above, my impression that there is a single root cause for a few of our current issues. I wrote up my thoughts here: #104. Feedback welcome. :-)

erdlet: Hey all,
how do we want to proceed with this PR now? Is #104 targeted before the upcoming release or afterwards? 

My problem is, that the PR related to the CSRF handling for other HTTP methods than POST is based on this branch, because I changed the test package structure here. If we want to this `ExceptionMapper` issue with #104 , this is OK for me, but I'd need to rebase the other PR ;)

chkal: I would love to address #104 before the upcoming release to be honest. I'm planning to spend some time tomorrow to work on a prototype and provide some feedback about the next steps afterwards.

However, I think if we can fix the underlying issues with #104, this PR request isn't needed anymore. So maybe rebasing the other PR against master would be a good way to proceed?

erdlet: :+1: I think #104 is the better way too. Unfortunately I haven't much time at the moment, so I can't help with #104 a lot. But I'm really looking forward to your prototype.

I'm going to rebase the other branch and pick the commit where I changed the test directory layout (I'd need to do it anyway) and close this PR afterwards.

erdlet:
Closed in favor of the solution proposed in #104 

Rebased #90 against master to be able to close this PR.

core/src/main/java/org/eclipse/krazo/core/ViewResponseFilter.java
chkal: > Unfortunately I haven't much time at the moment, so I can't help with #104 a lot. But I'm really looking forward to your prototype.

No problem! You helped us a lot in the last weeks. :beers:

Issue #70, opened 2019-05-13
Testsuite failures on wildly


gtudan:
Our test suite has a couple of tests that fail on wildly:

- [ ] The Asciidoctor-ext has class loading issues on wildly. JRuby doesn't find the Asciidoctor module, which is a [known issue](https://github.com/asciidoctor/asciidoctorj/blob/26641dbe1510a83fbb9c5cab3974a8736b385bdd/README.adoc#running-asciidoctorj-on-wildfly-as) at Asciidoctorj
- [ ] JSR-223 fails for python. Probably something similar as in the jruby/asciidoctor issue above. 
- [x] The Jetbrick test fails and I'm not sure why. The project documentation and issue tracker is in mandarin(?), so looking for known issues is a bit difficult.

Most of those failures are probably issues with the test suite and not bugs in the extensions. But they point out potential pitfalls that should be documented somewhere (readme of the extension?).

gtudan:
Strange - the jetbrick test works locally. Might be something on Travis.

The Jython-issue is a little odd as well. The engine get's detected by the ScriptManager, but isn't registered correctly: the SPI get's picked up, but there are no bindings for the names or extensions. The initialization is run (I can see some log output), but there's no error explaining the issue.

gtudan:
I'm afraid getting the asciidoctorj-example working in the current test suite would be very hard. We would need to patch wildfly to add asciidoctor as module: https://github.com/asciidoctor/asciidoctorj/blob/26641dbe1510a83fbb9c5cab3974a8736b385bdd/README.adoc#running-asciidoctorj-on-wildfly-as

The guys at asciidoctor did it some time ago with a dedicated docker image: https://asciidoctor.org/news/2015/03/04/docker-asciidoctorj-announcement/
Today they use Gradle to patch wildfly: https://github.com/asciidoctor/asciidoctorj/tree/master/asciidoctorj-wildfly-integration-test

I'm not sure if it is worth the effort to build something like this for krazo. Should we exclude the Asciidoctor-Test from the wildfly-testrun?

chkal:
> I'm not sure if it is worth the effort to build something like this for krazo. Should we exclude the Asciidoctor-Test from the wildfly-testrun?

Hmmm. Too bad. I agree that it would be fine to skip the AsciiDoc test in this case. A few years ago I did something similar in the rewrite project. 

See:

[...]

PR #88, opened 2019-07-29
Disable AsciiDocIT on wildfly.

erdlet:
see: #70

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -277,6 +277,9 @@
                                 <arquillian.launch>wildfly</arquillian.launch>
                                 <testsuite.profile>testsuite-wildfly</testsuite.profile>
                             </systemProperties>
+                            <excludedGroups>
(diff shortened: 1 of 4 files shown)


erdlet:
Always happy to help :)

chkal:
Thanks a lot, looks great. As mentioned in the other pull request, I would like to wait a bit with merging this, so we can do the bulk-update of the copyright headers before.

One minor thing: Could you perhaps add a comment to the `@Category(IgnoreOnWildfly.class)` lines explaining why we are ignoring these tests on Wildfly? Just to make sure that we remember the reason in a few years. :laughing:

erdlet:
Sure makes sense. I'll add comments on these tests :)

chkal:
Unfortunately getting approval for the update of the copyright headers takes longer than expected. So I'll start to merge a few of the pending pull requests and rebase my branch for the header update.

Sorry for the delay... :disappointed:

PR #90, opened 2019-09-14
89 extend csrf protection

erdlet:


erdlet:
This PR is based on the branch of #87, because I changed the directory structure of CSRF tests there. Please only review the last commit `c3ae5d4` for this PR. Thanks!

chkal:
Thanks a lot! I hope that we can merge the open pull requests soon, so we can proceed with working on Krazo.

BTW: The dependency check for this PR failed with:

```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:5.1.0:aggregate (default-cli) on project krazo-parent: 
[ERROR] 
[ERROR] One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '7.0': 
[ERROR] 
[ERROR] commons-beanutils-1.9.3.jar: CVE-2019-10086
[ERROR] 
[ERROR] See the dependency-check report for more details.
```

chkal:
Ah, sorry, I just saw that this dependency wasn't introduced by the PR. I guess this is some new CVE for an existing dependency.

erdlet:
Saw that too yesterday but had no time to fix it, sorry. I'm on vacation
until next Saturday, so I'll update it after my return.

Christian Kaltepoth <notifications@github.com> schrieb am So., 15. Sep.
2019, 11:27:

chkal:
No need to say sorry. Looks like this is actually nothing caused by your PR.

Enjoy your vacation.

erdlet:
Rebased branch against master as talked about in #87 to be able to close #87 without merge.

core/src/main/java/org/eclipse/krazo/security/CsrfValidateFilter.java
```diff
@@ -105,12 +110,14 @@ public void filter(ContainerRequestContext context) throws IOException {
...
-            final List<String> tokenValues = form.asMap().get(token.getParamName());
+            final List<String> tokenValues = form.asMap()
+                .get(token.getParamName());
             if (tokenValues == null || tokenValues.isEmpty()) {
                 throw new CsrfValidationException(messages.get("CsrfFailed", "missing field"));
             }
 
-            if (!token.getValue().equals(tokenValues.get(0))) {
+            if (!token.getValue()
+                .equals(tokenValues.get(0))) {
```
chkal: There are some weird code format changes included in the PR. Could you perhaps keep the original formatting?

erdlet: Yes I can revert it and keep an eye on it in the next PRs. Do we have somewhere a common formatter for Krazo?

chkal: AFAIK there is no code format profile yet. AFAIK the original code used the IntelliJ IDEA default configuration. But I can check that. I'll create an issue for tracking this. Thanks for bringing this up.

chkal:
More reviewers welcome! :wink:

testsuite/src/main/resources/csrf/methods/views/csrf-ok-post.jsp
```diff
@@ -0,0 +1,18 @@
...
+<html>
+<head>
+    <title>CSRF Custom Exception Test</title>
+</head>
+<body>
+<h1>CSRF Custom Exception Test</h1>
+<form id="form" action="${mvc.uri('doPost')}" method="post" accept-charset="utf-8">
+
+    <input type="hidden" name="_method" value="DELETE">
+
```
Daniel-Dos: Here is Delete ?

erdlet: Oh yes, good catch, thanks! Seems I forgot to change it in this form and the PATCH form.

erdlet: Should be fixed now :+1:

erdlet:
I also applied the new copyright header to the new classes. Maybe we can add a Maven plugin to check or apply the header automatically, so we can't forget it.

chkal:
> Maybe we can add a Maven plugin to check or apply the header automatically, so we can't forget it.

We had such a plugin in the past. But it checked against a static copyright header template, which won't work for us if we update the copyrights correctly after file modifications.

I already thought about added a simple shell script based check to the Travis config. Maybe this is enough? Or does anyone know about a plugin which would work for us?

gtudan:
> I already thought about added a simple shell script based check to the Travis config. Maybe this is enough? Or does anyone know about a plugin which would work for us?

The Maven-License-Plugin is quiet flexible flexible and could at least ensure that the license, organization and inception year is correct:

https://www.mojohaus.org/license-maven-plugin/check-file-header-mojo.html

erdlet:
Maybe we should discuss this in an extra issue?

chkal:
> Maybe we should discuss this in an extra issue?

I agree! Let's file a separate issue for that. This is quite off-topic here. :-)

chkal:
I just merged the PR!

@erdlet Thanks a lot for your help!

PR #94, opened 2019-10-07
Disable processing of view-annotations if controller does not complete

### LINKED ISSUE #91: CSRF-Validation does not work with @View annotation
After lots of debugging for #27 I found that `CsrfProtected` does not work in Combination with `@View`.

The CsrfValdiateFilter throws an exception as expected. This will skip the usual controller handling and trigger the CsrfExceptionMapper. There, the exception will be mapped to a response with corresponding Status and Reason Phrase, but without an entity. Everything is fine until here.

But then there is a special handling in the `ViewResponseFilter` that checks for calls with an empty entity on controller methods with the `@View` annotation. 

https://github.com/eclipse-ee4j/krazo/blob/master/core/src/main/java/org/eclipse/krazo/core/ViewResponseFilter.java#L133

It will kick in and render the page despite any previous Exceptions, even if the controller never executed. 

Consider this controller: 
```java
    @POST
    @CsrfProtected
    @View("redirect:/csrf/ok")
    public void postForm(@FormParam("greeting") String greeting) {
        
    }
```

This issue is not restricted to CsrfExceptions - it can happen on any ExceptionsMappers that don't return an entity in combination with the `@View`-Annotation.

What can we do about it?
An easy fix for CSRF-Validation would be to make the ExceptionMapper return a default-view. Compliant implementation must support JSP, so we could default to this as engine. We would need to think about how to inject our built-in-views, but it could be done.

The other approach would be to somehow detect that an exception handler kicked in inside the `ViewResponseFilter` and disable the view annotation processing.

I would prefer to push the handling of the `@View` annotation further down towards the controller. Maybe we can decorate the controller, so that it reads the annotation and writes the view from there to the entity if (and only if) it returns without an exception? A CDI-Interceptor running before the `AroundControllerInterceptor` could do the trick.

Here's a failing unit-test demonstrating the issue:
https://github.com/gtudan/krazo/commit/c7da79473f18bcb0e769dddc7ee9a5aa7deabf0c

gtudan:
The processing of the view annotations is disables by setting an attribute in the aroundControllerInterceptor if and only if the controller ran and completed without an exception.

Fixes #91 where views configured by a view annotation would still be rendered if an exception handler kicked in.

Sorry about the amount of change in the ViewResponseFilter - I had to catch some edge cases and get the class under test.

The test requires a lot of dirty mocking, but I did not want to refactor even more stuff to keep the PR readable.

### DIFF
--- a/core/src/main/java/org/eclipse/krazo/cdi/KrazoCdiExtension.java
+++ b/core/src/main/java/org/eclipse/krazo/cdi/KrazoCdiExtension.java
@@ -102,7 +102,7 @@ public class KrazoCdiExtension implements Extension {
                 ModelsImpl.class,
                 ViewableWriter.class,
                 ViewResponseFilter.class,
-                
+
                 // lifecycle
                 EventDispatcher.class,
                 RequestLifecycle.class,
@@ -213,5 +213,14 @@ public class KrazoCdiExtension implements Extension {
      */
     public static synchronized boolean isEventObserved(Class<? extends MvcEvent> eventType) {
         return observedEvents == null ? false : observedEvents.contains(eventType);
+    }
+
+    /**
+     * Clears all observed events
+     * <p>
+     * This is used for testing
+     */
+    public static synchronized void clearEventsObserved() {
+        observedEvents = null;
     }
 }
--- a/core/src/main/java/org/eclipse/krazo/core/RequestAttributes.java
+++ b/core/src/main/java/org/eclipse/krazo/core/RequestAttributes.java
@@ -1,0 +1,9 @@
+package org.eclipse.krazo.core;
+
+/**
+ * @author Gregor Tudan
+ */
+public enum RequestAttributes {
+
+    CONTROLLER_EXECUTED
+}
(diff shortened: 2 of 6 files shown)


gtudan:
@chkal I still have to find out why the test suite is failing on the inheritance tests. I might need some help here as the rules for reading the annotations haven't really changed. Could that be an exception in the test suite that was hidden by the broken exception handling?

chkal:
> I still have to find out why the test suite is failing on the inheritance tests. I might need some help here as the rules for reading the annotations haven't really changed. Could that be an exception in the test suite that was hidden by the broken exception handling?

The logs don't reveal much about the underlying issue. That's weird.

Maybe we should first implement the changes discussed in #104 and see if this issue still appears. And if it does, I can do some debugging with the TCK. This may really be caused some kind of hidden root cause.

chkal:
I just merged #106. Should we close this one or does it make sense to at least merge the test contained in this one?

gtudan:
Never mind about the tests, they are rather hacky due to the many mocks anyway. Let’s close this.

PR #98, opened 2019-10-12
Fixed copyright line in file headers

chkal:
This pull request fixes the copyright headers in the whole Krazo codebase. Basically it restores the original Oracle copyright lines in the code that was transferred back then and it unifies the headers to correctly indicate the copyright for all content added later.

I know that this is a very huge pull request. However, feel free to comment if you have any doubts.

### DIFF
--- a/testsuite/src/main/resources/jade/views/includes/footer.jade
+++ b/testsuite/src/main/resources/jade/views/includes/footer.jade
@@ -1,2 +1,2 @@
 #footer
-  p.footer Copyright (c) Ivar Grimstad
+  p.footer Copyright (c) Eclipse Krazo committers and contributors
--- a/core/findbugs-exclude.xml
+++ b/core/findbugs-exclude.xml
@@ -1,6 +1,7 @@
 <!--
 
-    Copyright © 2017, 2018 Ivar Grimstad
+    Copyright (c) 2014-2015 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2018, 2019 Eclipse Krazo committers and contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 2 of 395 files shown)


chkal:
Ups. Broken build. Looks like we need to get in #99 before.

chkal:
Rebased against master to get the fix from #99 in.

testsuite/src/main/resources/jade/views/includes/footer.jade
```diff
@@ -1,2 +1,2 @@
 #footer
-  p.footer Copyright (c) Ivar Grimstad
\ No newline at end of file
+  p.footer Copyright (c) Eclipse Krazo committers and contributors
```
gtudan: The integration tests checks a copyright header? That's - creative... 😄

But it reminds me that most of the templates from the test suite and the the examples don't have a copyright header at all - is that a problem?

chkal: No, I think that's not a problem at all. I guess adding copyright headers source files is required, but for most other files I don't think that this is really necessary.

chkal: The jade template is just a weird exception to that. And the fact that the test checks the content of the footer actually caused some "surprising results" when I changed the headers the first time. :laughing:

chkal:
I guess we shouldn't wait too long with merging this one. So I'll do it now.

Everyone is welcome to review the PR nevertheless. If anything is not correct in your opinion, please feel free to comment on this PR and we can discuss what needs to be changed.

testsuite/src/main/resources/jade/views/includes/footer.jade
lefloh: It doesn't check the copyright. It checks if a jade`include` is working.

Issue #97, opened 2019-10-09
Allow building with OpenJDK > 8


gtudan:
I would like to be able to compile the project with more recent JDK-Versions. The Oracle JDK8 is no longer supported, so we shouldn‘t require developers to stick with it.

The build can be kept backwards-compatible with earlier Java versions by using the -release flag.

We already did this for the the MVC-Spec (see its pom-File).

chkal:
Big +1 for this one!

mthmulders:
Good one! Maybe we could use a [Build Matrix](https://docs.travis-ci.com/user/build-matrix/) to build on multiple Java versions at the same time. What versions do we want to build on? I'd guess at this point 8, 11 and 13 (with 11 being the latest LTS and 13 being the latest regular release)...

chkal:
Maybe we should change `oraclejdk8` to `openjdk8` in a first step?

Basically I like the idea of running the build with multiple JDK version, but the build matrix is already a bit complex, as we run the integration tests and the TCK against multiple container profiles.

mthmulders:
Switching to OpenJDK would be a good (and simple) first step. I'll provide a PR for that, it's a quick win.

I think the next thing we could do is start _building_ Krazo with multiple Java versions. The last step would be to run the TCK against multiple containers running multiple versions of Java. That last part is indeed going to be a complex one. I just don't know yet if we could separate the phases of the build like that...

PR #100, opened 2019-10-12
Switch from Oracle JDK to OpenJDK

mthmulders:
As discussed in #97, this switches the build from Oracle JDK to OpenJDK. The Java version stays the same, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 sudo: required
-jdk: oraclejdk8
+jdk: openjdk8
 
 services:
 - docker


chkal:
@mthmulders Thanks a lot! :-)

I created #99 a few minutes earlier. Looks like we both wanted to fix this at the same time. :-)

However, the ECA check if failing for this one. Do you have an Eclipse account and signed the ECA?

I'll merge #99 instead to get this fix in quickly. Thanks however for helping with this. :beers:

chkal:
Closed in favor of #99.

mthmulders:
I wasn't aware of the ECA until I noticed the check failed... I think I have that one set up now, so let's hope the next PR will be approved :-).

chkal:
Yeah, the Eclipse Foundation takes these kinds of agreements very seriously. But once set up, everything should work without problems. You just have to use a signed-of header in your commit.

PR #101, opened 2019-10-12
Allow building with Java 9 and above

### LINKED ISSUE #97: Allow building with OpenJDK > 8

gtudan:
This detects Java 9 and above and sets the `maven.compiler.release` property.
This makes the Java compiler generate Java 8 byte code, so we keep backwards compatibility with older Java versions.

Closes #97

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ jobs:
   include:
     - stage: build
       script: mvn -B -V -DskipTests -DperformRelease -Dgpg.skip clean install
+    - stage: build
+      script: mvn -B -V -DskipTests clean install
+      jdk: openjdk11
     - stage: test
       env: TYPE=glassfish-module
       script: .travis/tests.sh ${TYPE}
--- a/pom.xml
+++ b/pom.xml
@@ -250,7 +250,15 @@
     </build>
 
     <profiles>
-
+        <profile>
+            <id>jdk9plus</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>8</maven.compiler.release>
+            </properties>
+        </profile>
         <profile>
             <id>release</id>
             <activation>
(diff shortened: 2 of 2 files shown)


chkal:
Hmmm. Doesn't work yet:

```
[INFO] -------------------------------------------------------------
[WARNING] COMPILATION WARNING : 
[INFO] -------------------------------------------------------------
[WARNING] bootstrap class path not set in conjunction with -source 8
[INFO] 1 warning
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] warnings found and -Werror specified
[INFO] 1 error
[INFO] -------------------------------------------------------------
```

IMO it is not super important to keep `-Werror`, but maybe getting rid of the warning is simpler?

chkal:
Looks like we can only get rid of the warning if we reference the `rt.jar` from that specific older JDK version... :disappointed: 

So maybe removing `-Werror` is the way to go?

gtudan:
Sorry about that, I was a bit in a rush when committing this.
We could install both openjdk8 and 11 using travis jdk_switcher, but there is no environment option for setting the bootstrap class path. That makes it hard to configure only for travis.

Disabling the warnings altogether sounds a bit drastic - how about disabling only the warnings about compiler options `-Xlint:-options`? I updated the PR.

chkal:
> Disabling the warnings altogether sounds a bit drastic

No, my proposal was not to disable warnings, but to don't let the compiler fail for warnings. That's what `-Werror` is about AFAIK.

> how about disabling only the warnings about compiler options `-Xlint:-options`?

+1. Sounds good. Maybe even better than removing `-Werror`.

Let's merge this one.

PR #103, opened 2019-10-13
run the unit tests in travis

gtudan:
I just noticed that we don't execute the unit tests during the travis build. All stages have `-DskipTests`.

Here's a failing test. The build is still green.
https://github.com/gtudan/krazo/commit/3ac2c34899efe47c1cdc63c6763a6435713b380b

Let's add a stage that runs them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ jobs:
     - stage: build
       script: mvn -B -V -DskipTests -DperformRelease -Dgpg.skip clean install
     - stage: test
+      script: mvn -B -V test
+    - stage: test
       env: TYPE=glassfish-module
       script: .travis/tests.sh ${TYPE}
     - stage: test


chkal:
We could also run unit test as part of the build phase. AFAIK we only unit tests in the core module at the moment. This way the test phase would only be for long-running integration tests. But I'm not strong on this one. Thoughts?

erdlet:
I like Gregors solution, because the build process is separated into one task per concern. This makes the debugging easier than the other way, when we would have a combined compile / unit-test build step.

chkal:
> I like Gregors solution, because the build process is separated into one task per concern. This makes the debugging easier than the other way, when we would have a combined compile / unit-test build step.

That's definitely a +1 for this config. But I guess it is all a matter of taste.

@erdlet BTW: Feel free to review and approve any PR if you like. Everyone can review and approve PRs and more feedback is always welcome.

gtudan:
I don't have a strong opinion on this - on the one hand our test stage is starting to get a little bloated and maybe we should not overcomplicate things. On the other hand, separating the different concerns (build and test) from one-another seems like the right thing to. But it is a matter of taste... 🤷‍♂

chkal:
I also don't have a super strong opinion. So let's merge this one now. We can change the setup later if we want...

Issue #104, opened 2019-10-14
Optimize rendering pipeline for the non-happy-path


chkal:
This issue aims to be some kind of umbrella discussion about a category of issues which is about the non-happy-path of an MVC request. So basically it is about what to do in case of exceptions and other scenarios.

Related issues:
  * CSRF validation vs. `@View` (issue #91 + PR #94)
  * Custom exception mappers problem (issue #34, PR #34)
  * Maybe others?

I'll try to summarize what my understanding of the underlying issue is. But I may be wrong with my following analysis. To be honest, I didn't find much time in the last weeks to dig into these issues. So please speak up if my statements don't make sense to you.

In a nutshell our rendering pipeline works like this (for the happy path):

  1. The controller is invoked and returns a view name (or `null` for `@View`)
  2. The `ViewResponseFilter` kicks in, checks the controller's return value (and `@View`) and replaces the response entity with a `Viewable` containing all the required information for rendering.
  3. The `ViewableWriter` then is called by the JAX-RS implementation to actually render the view.

This works fine on the happy path. Now if exceptions occur (like for a failed CSRF validation), things are getting more complicated. It looks like in this case step (2) is executed which leads to all sorts of problems:

  * An exception mapper may want to create a plain text response but the filter will try to render a corresponding view (see #34).
  * The filter will look for `@View` annotations if a mapper created a null-entity (see #91)
  * I'm pretty sure there are more issues...

So I'm wondering whether we can find a really simple solution for all of this. What if we just check in `ViewResponseFilter` if we are on the happy path (controller was executed successfully) do nothing otherwise. What does this mean? IMO all the problems mentioned above are gone. Exception mappers can do their work as they can without Krazo in the classpath. 

The only remaining question: what if the mapper wants to render a view? In this case he could simply use a `Viewable` for the entity in which case the request processing will continue as usual with step (3). For now this would be a Krazo-specific feature as `Viewable` is not part of the API anymore. Actually we removed it back then because the spec was very vague about many specific aspects. But I'm absolutely fine with adding it back to the spec for the next spec release.

@gtudan @erdlet Thoughts? As mentioned above, maybe I'm missing something obvious here.

chkal:
BTW: I think that #94 is a real good basis for the fix I'm suggesting here.

gtudan:
I like the idea, especially for the simplicity. 

[...]

Issue #34, opened 2019-01-18
Custom CSRF exception mappers are always processed by Krazo


chkal:
MVC allows to define custom exception mappers for `CsrfValidationExceptions`:

```java
@Provider
public class MyMapper implements ExceptionMapper<CsrfValidationException> {

    @Override
    public Response toResponse(CsrfValidationException exception) {
        return Response.status(403)
                .type(MediaType.TEXT_PLAIN)
                .entity("Invalid CSRF token")
                .build();
    }

}
```

However, using this mapper currently fails with:

```java
javax.ws.rs.ServerErrorException: Unable to find suitable view engine for 'org.eclipse.krazo.engine.Viewable@5f01e840'
	at org.eclipse.krazo.core.ViewableWriter.writeTo(ViewableWriter.java:135)
	at org.eclipse.krazo.core.ViewableWriter.writeTo(ViewableWriter.java:77)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.invokeWriteTo(WriterInterceptorExecutor.java:266)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.aroundWriteTo(WriterInterceptorExecutor.java:251)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor.proceed(WriterInterceptorExecutor.java:163)
	at org.glassfish.jersey.server.internal.JsonWithPaddingInterceptor.aroundWriteTo(JsonWithPaddingInterceptor.java:109)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor.proceed(WriterInterceptorExecutor.java:163)
	at org.glassfish.jersey.server.internal.MappableExceptionWrapperInterceptor.aroundWriteTo(MappableExceptionWrapperInterceptor.java:85)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor.proceed(WriterInterceptorExecutor.java:163)
	at org.glassfish.jersey.message.internal.MessageBodyFactory.writeTo(MessageBodyFactory.java:1135)
	at org.glassfish.jersey.server.ServerRuntime$Responder.writeResponse(ServerRuntime.java:662)
	at org.glassfish.jersey.server.ServerRuntime$Responder.processResponse(ServerRuntime.java:395)
	at org.glassfish.jersey.server.ServerRuntime$Responder.process(ServerRuntime.java:441)
[...]

Issue #29, opened 2018-12-07
Inconsistent behavior of annotations on Glassfish and Wildfly


mvcbot:
<a href="https://github.com/gtudan"><img src="https://avatars1.githubusercontent.com/u/419425?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [gtudan](https://github.com/gtudan)**
_Tuesday Dec 04, 2018 at 06:53 GMT_
_Originally opened as https://github.com/mvc-spec/ozark/issues/220_

----

Wildfly seems to behave different than Glassfish when inheriting annotations.

The "Annotations"-Example-App contains a test case for that: https://github.com/mvc-spec/ozark/blob/b94d83ff85a9267f1ae643650cf0f461a06f9021/examples/annotations/src/test/java/org/mvcspec/ozark/test/annotations/AnnotationsIT.java#L75

Glassfish will return an Error (HTTP 500) for this test, while wildfly renders the view with a HTTP 200. I guess WF inherits the annotation from the BaseController. This violates the annotation inheritance rules defined in part 6.4 of the MVC-Spec.

mvcbot:
<a href="https://github.com/chkal"><img src="https://avatars2.githubusercontent.com/u/159245?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [chkal](https://github.com/chkal)**
_Tuesday Dec 04, 2018 at 07:55 GMT_

----

Just a quick note: My current guess is that this issue is related to the typical problems we get with CDI proxies. Especially the inheritance rules get very tricky if additional CDI proxy subclasses are involved.

dmaidaniuk:
@chkal, 
I will take a look at this issue and write here what I found.

chkal:
@dmaidaniuk Awesome! Looking forward to reading about your analysis.

chkal:
Would be interesting to know if this is still an issue after #106 has been merged. Maybe we could migrate the annotations sample app to the new testsuite so we can run it against different containers more easily.

erdlet:
Sounds like an good idea. I have some time today, so I can migrate it and create a PR.

chkal:
@erdlet Awesome! Thanks a lot!

[...]

PR #106, opened 2019-10-21
Run ViewResponseFilter only if controller was executed successfully

chkal:
This pull requests contains the refactoring described in #104. Actually the change is quite simple. The request-scoped`RequestLifecycle` class just tracks if the controller was invoked successfully and the `VewiResponseFilter` only runs if this is the case. 

The PR contains integration tests which checks 3 different scenarios:

  * If the controller execution is successful, the request is processed just like before.
  * If the controller throws an exception and an exception mapper creates a plain text response, this must work exactly like in a plain JAX-RS application (and NOT trigger a view engine lookup)
  * If the controller throws an exception and an exception mapper wants to render an MVC view, it can simple use a `Viewable` as the entity.

I guess this should cover all the relevant cases. The integration test actually reproduces the exact case described in #34. I don't have an explicit coverage for #91 yet, but I'm pretty sure that this is fixed as well.

The PR also contains another fix: The `AroundControllerInterceptor` wasn't detecting PATCH methods and was not implementing the MVC / JAX-RS annotation inheritance rules correctly. This caused test failures after my refactoring. The fix was easy, because there is already a utility method which checks if a method is a controller method.

Let me know what you think.

### DIFF
--- a/core/src/main/java/org/eclipse/krazo/util/ControllerUtils.java
+++ b/core/src/main/java/org/eclipse/krazo/util/ControllerUtils.java
@@ -64,8 +64,8 @@ public final class ControllerUtils {
         if (hasDeclaredRequestMethodAnnotation(method)) {
             return true;
         }
-        // inheritance disabled if other MVC or JAX-RS annotations found
-        if (AnnotationUtils.hasMvcOrJaxrsAnnotations(method)) {
+        // inheritance disabled if other JAX-RS annotations found
+        if (AnnotationUtils.hasJaxrsAnnotations(method)) {
             return false;
         }
         // check all super classes
--- a/core/src/main/java/org/eclipse/krazo/cdi/types/AnnotatedTypeProcessor.java
+++ b/core/src/main/java/org/eclipse/krazo/cdi/types/AnnotatedTypeProcessor.java
@@ -19,11 +19,10 @@ package org.eclipse.krazo.cdi.types;
 
 import org.eclipse.krazo.binding.validate.ValidationInterceptorBinding;
 import org.eclipse.krazo.cdi.AroundController;
+import org.eclipse.krazo.util.ControllerUtils;
 
 import javax.enterprise.inject.spi.AnnotatedMethod;
 import javax.enterprise.inject.spi.AnnotatedType;
-import javax.mvc.Controller;
-import javax.ws.rs.*;
 import java.lang.annotation.Annotation;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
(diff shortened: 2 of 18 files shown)


erdlet:
```
Tests in error: 

  AnnotationsIT.testNoOverrideMvc:80 NullPointer
```

Only one error, but it occured also in #94.

chkal:
> Even tough the build fails, I approve the PR because the build fail is in my opinion nothing which is directly caused by this change.

Unfortunately I'm pretty sure that these failures reveal some hidden bug in Krazo. Because with my patch the view rendering only works if the CDI interceptors are applied correctly. And it looks like that this isn't the case for these weird corner cases related to annotation inheritance. :disappointed: 

> I think you'd need to add me to the project as contributor or something similar. Now I can approve also, but it isn't counted into the PR checks. But that's OK for me.

Unfortunately we cannot change this directly, because we don't have admin permissions for the repo. However, of course you are always welcome to review ANY pull request. :+1:

gtudan:
> Unfortunately I'm pretty sure that these failures reveal some hidden bug in Krazo. Because with my patch the view rendering only works if the CDI interceptors are applied correctly. And it looks like that this isn't the case for these weird corner cases related to annotation inheritance.

That rings a bell... remember #29? I think I saw something similar there.

chkal:
> That rings a bell... remember #29? I think I saw something similar there.

Yeah, possible. The inheritance rules are tricky to implement.

chkal:
Looks like all the tests are green now. There were two major issues which caused quite some pain:
  * There was a corner case in the annotation inheritance implementation which caused one test to fail. See af64422eb7017dd0b9a59a1beeade321a3e381dc for details.
  * A few of the old tests checked that `@View` is considered if a controller methods fails with an exception. This doesn't make much sense to me, because this `@View` would also be used for the success case. I adjusted these old tests in 450657d68151423e1a5aa8705d7846d8b6c979b0 and 61cb5615f7bc9a1c6f879ab108af999de55f2103

core/src/main/java/org/eclipse/krazo/util/ControllerUtils.java
```diff
@@ -65,7 +65,7 @@ static boolean isRequestMethod(Method method) {
             return true;
         }
         // inheritance disabled if other MVC or JAX-RS annotations found
```
erdlet: We should remove this comment as it is wrong now. The method name speaks for itself here.

chkal: Good catch. I adjusted the comment.

PR #108, opened 2019-10-23
Migrate AnnotationIT from examples to testsuite.

erdlet:
see: #29

To verify if the issue still occurs after #106, the example is migrated to
the integration test suite.


erdlet:
@chkal @gtudan 
Would be great if you or someone else could review this. Unfortunately, I can't request a review :/

core/src/main/java/org/eclipse/krazo/cdi/AroundControllerInterceptor.java
```diff
@@ -34,14 +36,17 @@
...
+
     private static final long serialVersionUID = -5804986456381504613L;
 
     @Inject
     private RequestLifecycle requestLifecycle;
 
     @AroundInvoke
     public Object validateMethodInvocation(InvocationContext ctx) throws Exception {
-        return requestLifecycle.aroundController(ctx::proceed);
+        LOGGER.log(Level.FINE, "Invocing method: {0}#{1}", new Object[] { ctx.getMethod().getDeclaringClass().getName(), ctx.getMethod().getName() });
```
chkal: Typo: `Invocing` -> `Invoking`

core/src/main/java/org/eclipse/krazo/lifecycle/RequestLifecycle.java
```diff
@@ -43,6 +44,8 @@
 
     private boolean controllerExecuted = false;
 
+    private Method invokedMethod;
```
chkal: Let's call this `controllerMethod` instead. I think that's a better fit.

testsuite/src/test/java/org/eclipse/krazo/test/annotations/AnnotationsIT.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Test
+    public void testNoView() throws Exception {
+        final HtmlPage page = webClient.getPage(baseURL + "resources/annotations/no_view");
+        final Iterator<HtmlElement> it = page.getDocumentElement()
+                .getElementsByTagName("h1")
+                .iterator();
+        assertTrue(it.next()
+                           .asText()
```
chkal: Could you remove those line breaks? I guess this was the automatic code formatter?

core/src/main/java/org/eclipse/krazo/lifecycle/RequestLifecycle.java
erdlet: I'd name it `invokedControllerMethod`, as the name `controllerMethod` is already taken by the `Callable` in `aroundController`. But maybe we find a better name for the current `controllerMethod` field?

chkal: Ah, I see. Maybe the `aroundController` method could look like this:

```
public Object aroundController(Method method, Callable<Object> invocation) .. {
```

I know that `method` is a bit generic, but maybe ok for this small method. Then we could call the field `controllerMethod` or `invokedControllerMethod`.

core/src/main/java/org/eclipse/krazo/lifecycle/RequestLifecycle.java
```diff
@@ -51,12 +54,12 @@ public void beforeAll(ContainerRequestContext context) {
...
     }
 
-    public Object aroundController(Callable<Object> controllerMethod) throws Exception {
+    public Object aroundController(final Method invokedMethod, Callable<Object> controllerMethod) throws Exception {
 
         eventDispatcher.fireBeforeControllerEvent();
         try {
-
             Object result = controllerMethod.call();
+            this.invokedMethod = invokedMethod;
```
chkal: I'm wondering if we should move this line up, so that we have access to `invokedMethod` before the controller is invoked and maybe even before the `BeforeController` events are fired. In this case we should prefer to call the field `controllerMethod` and not `invokedControllerMethod` because it is available before the controller has been invoked.

core/src/main/java/org/eclipse/krazo/lifecycle/RequestLifecycle.java
erdlet: Sounds good, I'll change it this way

PR #109, opened 2019-10-24
Add custom resource loader for Velocity view engine

chkal:
This PR removes our dependency on `velocity-tools-view`. I added a custom resource loader which replaces the one we used from the old dependency. Our new custom implementation doesn't support any caching at the moment, but I don't think that this is a real problem. Especially because most of the other 3rd party view engines also don't support a caching mechanism.

I also updated Velocity Engine from 2.0 to 2.1. There is a [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21009) which was already approved

I was also able to remove the `<dependencyManagement>` entry we needed due to the outdated `commons-beanutils` dependency in Velocity Tools.

### DIFF
--- a/ext/velocity/pom.xml
+++ b/ext/velocity/pom.xml
@@ -31,9 +31,9 @@
     <name>Eclipse Krazo Velocity Extension</name>
     <dependencies>
         <dependency>
-            <groupId>org.apache.velocity.tools</groupId>
-            <artifactId>velocity-tools-view</artifactId>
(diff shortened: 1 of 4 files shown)


chkal:
> It looked more complex in the Velocity implementation.

Yeah, the `ResourceLoader` interface is a bit weird as it assumes file-like templates with `modifiedAt` dates, etc. But my implementation ignores this for now.

PR #114, opened 2019-11-12
update dependencies to CXF and RestEasy

Daniel-Dos:
this pr make update to CXF and RestEasy . 

Signed-off-by: Daniel Dias <daniel.dias.analistati@gmail.com>

### DIFF
--- a/cxf/pom.xml
+++ b/cxf/pom.xml
@@ -32,7 +32,7 @@
     <name>Eclipse Krazo CXF</name>
 
     <properties>
-        <cxf.version>3.3.1</cxf.version>
+        <cxf.version>3.3.4</cxf.version>
     </properties>
 
     <build>
--- a/resteasy/pom.xml
+++ b/resteasy/pom.xml
@@ -32,7 +32,7 @@
     <name>Eclipse Krazo RESTEasy</name>
 
     <properties>
-        <resteasy.version>3.6.2.Final</resteasy.version>
+        <resteasy.version>3.6.3.Final</resteasy.version>
     </properties>
 
     <build>


chkal:
Thanks!

Please note that we will need CQs for these. Although I'm not completely sure if this is required for patch releases. I'll check that.

Daniel-Dos:
hi 
how I make this ?

thanks :  )

chkal:
Creating CQs isn't an easy process. But the EF is currently improving this. See [this blog post](https://waynebeaton.wordpress.com/2019/10/09/revising-the-eclipse-ip-policy-third-party-content/) by @waynebeaton for details.

Therefore, I think we should delay the merge of this PR, as I'm planning to create the IP Log for the final release really soon. I guess updating the dependencies isn't that important, because both are `provided` scoped.

waynebeaton:
> Therefore, I think we should delay the merge of this PR, as I'm planning to create the IP Log for the final release really soon. 

CQs are not required for service releases of content that has already been approved.

> I guess updating the dependencies isn't that important, because both are `provided` scoped.

If the suggestion is that it isn't important because the consumer will get the newest version anyway, then you've misinterpreted the IP Policy. If you have reasonable expectation that a consumer will use a particular version of a library, then that library is subject to the IP Due Diligence Process. 

In this particular case IP Due Diligence Process says that you don't need a CQ.

More here: https://www.eclipse.org/projects/handbook/#ip-third-party-versions

chkal:
@waynebeaton Thanks a lot for the clarification. So I was actually wrong about my statement regarding the requirement of a CQ for these updates. Thanks a lot!

@Daniel-Dos In this case let's merge this one now!

Daniel-Dos:
thanks :   )

PR #116, opened 2019-11-19
Updated Java EE API with Jakarta EE version

chkal:
We currently don't have a CQ for the Java EE 7 dependency. And as we need Java/Jakarta EE 8 anyway, let's update this now.


erdlet:
Thanks for the updates. I'll remove the JEE 7 dependencies from the other issue.

PR #117, opened 2019-11-26
115 update bugfix versions

erdlet:


chkal:
@erdlet Thanks a lot for the PR.

For some reason the test suite is failing now. The only dependency affected that could cause this is the update of `graphene-webdriver`. I just restarted the build to make sure that this isn't a temporary failure.

erdlet:
Thanks a lot.
I'll check this dependency update in case it'll fail again.

Christian Kaltepoth <notifications@github.com> schrieb am Mi., 27. Nov.
2019, 19:13:

erdlet:
The root cause of this problem is `Caused by: java.lang.NoClassDefFoundError: org/openqa/selenium/interactions/internal/Locatable`.

This is the result of an old Selenium version which is managed by the `arquillian-drone-bom`, which seems to overwrite the version of the newer `graphene-webdriver`.

I'll revert this version upgrade and note the dependency between these dependencies.

chkal:
Great! Thanks! All tests are green now!

Issue #118, opened 2019-11-28
Double format problems when using <input type="number">


erdlet:
## Description
I recognized that the `NumberConverter` can't handle decimals from a number input field properly when using a non-english (or better: not-dot separated) locale.

Because the browser only sends dot-separated values to the server, the german-localed `NumberFormat` is not able to parse the value correct.

It would be great to find a solution for this, as the workaround, which is using a text input, is not the best solution from an semantic perspective (and UX perspective or accessibility perspective).

## Example
*Locale:* de-DE (via Accept-Language header)
*Input value:* 899,9 (HTML page & input is localed as german via MvcContext#getLocale)
*Request param:* "899.9"
*Output value:* 8999.0

## Reproduction
Simply take the current `DoubleConverterTest` and change the Locale in `ConversionTest` from `ENGLISH` to `GERMAN`.

## Workaround
Just use a text input field.

erdlet:
@chkal @gtudan what is your opinion on this (I saw, that you two worked a lot on the converters and tests). 

I think it's not that critical that we need to fix this before the release, but in the future it would be good to have a working decimal number input field because of the reasons mentioned above (I forgot to mention, that a number input field is ways better for mobile phones than a normal text field).

chkal:
Thanks for bringing this up.

I already ran into a similar issue a year ago in some other context:

Imagine a `<input type="number">` on a plain old HTML form:

* User 1
  * Uses a modern browser which supports HTML5 number inputs
  * The user enters `1,234` (German format)
  * The browser will submit: `1.234`
* User 2
  * Uses an older browser which doesn't support number inputs
  * The user enters `1,234` (German format)
  * The browser will fall back to standard `text` number inputs and submit: `1,234`

Now you have two outcomes. Either the result is `1234` or `1.234` depending on which locale you use and you cannot distinguish them.

:disappointed:

erdlet:
Did you find a solution in this other context? Or was the solution the workaround the same I use now by using textboxes?

It seems to me, that this is more an HTML issue because the number inputs are not really internationalizable :thinking:

gtudan:
Hi Tobias,

You are right, input type number has quiet a few known issues when it comes to internationalization. Take a look at the known issues tab on caniuse: https://caniuse.com/#feat=input-number
[...]

PR #123, opened 2019-12-15
118 extend converter registry

erdlet:


chkal:
Hey @erdlet. Awesome. Thanks a lot. I'll try to find some time for a review ASAP.

erdlet:
> Maybe we should use the ServiceLoader mechanism instead, which we already use for other SPIs?

I don't know the `ServiceLoader` mechanism very well, so maybe this is a better way to build this SPI. Using this would mean, that instead of maintaining and registering the converters in the `ConverterRegistry`, the `ConverterRegistry#lookup` method will call the `ServiceLoader` everytime it needs some converter?
As we have some time before merging this, I can check if we benefit from using the `ServiceLoader` here and change the implementation.

chkal:
> I don't know the `ServiceLoader` mechanism very well, so maybe this is a better way to build this SPI.

I personally like the ServiceLoader API, because it is a core JDK feature without any dependencies on frameworks like CDI, etc. But maybe that's a matter of taste. But we are already using it in various other places, so maybe this is a good reason to use it?

> Using this would mean, that instead of maintaining and registering the converters in the `ConverterRegistry`, the `ConverterRegistry#lookup` method will call the `ServiceLoader` everytime it needs some converter?

I don't think that we have to lookup the converters each time `lookup` is called. Especially because the converters available won't change at runtime. But we could use `ServiceLoader` in the registry's `@PostConstruct` method instead of hardcoding the default implementations. Looking up implementations using the `ServiceLoader` API is really easy and we alread do it here:

https://github.com/eclipse-ee4j/krazo/blob/f6915fd4ae75c0ab47979b3fde7bef776bda9921/core/src/main/java/org/eclipse/krazo/bootstrap/Initializer.java#L61-L64

erdlet:
> But we are already using it in various other places, so maybe this is a good reason to use it

This and the fact its a core JDK feature are good reasons enough for me to use the `ServiceLoader` instead of something selfmade based on some framework ;) 
Thanks for the `Initializer` example!

erdlet:
I changed the implementation, so the `ServiceLoader` is used now. I really like this approach, so thanks for giving me the hint to this API :)

core/src/main/java/org/eclipse/krazo/binding/convert/ConverterRegistry.java
```diff
@@ -37,24 +36,12 @@
...
-        register(new BooleanConverter());
+        converters.addAll(ServiceLoaders.list(MvcConverter.class));
     }
 
-    private void register(MvcConverter converter) {
-        converters.add(converter);
-    }
-
-    <T> MvcConverter<T> lookup(Class<T> rawType, Annotation[] annotations) {
+    public <T> MvcConverter<T> lookup(Class<T> rawType) {
```
chkal: I wonder if we should keep the annotations and also add them to `MvcConverter#supports()` and maybe even `MvcConverter#convert()`. Actually it was mistake that I didn't add it in the first version. I could imagine that this would be useful if you want to customize the conversion for a standard data type only for certain scenarios. Like for example if you want to add an annotation like `@ForceWireFormat`. Thoughts?

core/src/test/java/org/eclipse/krazo/binding/convert/ConverterRegistryTest.java
```diff
@@ -1,71 +0,0 @@
...
-import java.util.Arrays;
-import java.util.Collection;
-
-import static org.junit.Assert.*;
-
-/**
- * Asserts that converters for all supported types have been registered
- * @author Gregor Tudan
- */
-@RunWith(Parameterized.class)
```
chkal: Does it make sense to keep this one?

core/src/main/java/org/eclipse/krazo/binding/convert/ConverterRegistry.java
erdlet: Sure, we can re-add it. I didn't saw any usage at the moment, so I removed it. But maybe we should re-add it in a separate PR, so this one doesn't contain too much new features.

core/src/test/java/org/eclipse/krazo/binding/convert/ConverterRegistryTest.java
erdlet: Nope, I just forgot to delete the test. All the tests from this class are moved to the integration test.

core/src/main/java/org/eclipse/krazo/binding/convert/ConverterRegistry.java
chkal: Sure, I agree, let's do it in a subsequent pull request.

core/src/test/java/org/eclipse/krazo/binding/convert/ConverterRegistryTest.java
chkal: Ah, I see. Sorry, I missed that. :laughing:

chkal:
I'll keep this one open for 1-2 days. Maybe we get more feedback and/or reviews.

erdlet:
I created #131 for the extension we discussed in one of the code comments. This way, #118 can be closed after this PR and gets not bloated with more features, as the original problem can be solved with the current solution.

PR #129, opened 2019-12-30
Update TCK version to avoid dependency problems during build

erdlet:
At the moment, our build isn't able to be executed with `mvn install` without having the TCK installed locally. To avoid this, I set the version to the currently released TCK artifact, so this problem doesn't occur anymore.

### DIFF
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -33,7 +33,7 @@
 
     <properties>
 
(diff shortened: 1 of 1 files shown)


erdlet:
Seems I have an old `master` on my local repo. So I close this. Sorry ;)

chkal:
At first, I was a bit confused, because I remembered that I already changed the version of the TCK. :laughing:

Issue #134, opened 2020-01-10
Remove Arquillian Drone from the testsuite


chkal:
Arquillian Drone depends on an ancient HTMLUnit version and is causing a lot of trouble (see https://github.com/eclipse-ee4j/krazo/issues/115#issuecomment-572563298).

We should remove Drone and use HTMLUnit directly as we already do it for the TCK.

chkal:
Fixed via #135

PR #135, opened 2020-01-10
Replace arquillian-drone with plain htmlunit

erdlet:
see: #134

Because arquillian-drone relies on outdated versions of e.g. htmlunit, the tests use now
plain htmlunit and drone is removed

### DIFF
--- a/testsuite/src/test/java/org/eclipse/krazo/test/ext/Jsr223IT.java
+++ b/testsuite/src/test/java/org/eclipse/krazo/test/ext/Jsr223IT.java
@@ -81,6 +81,6 @@ public class Jsr223IT {
     @Category(IgnoreOnWildfly.class)
     public void testJython() throws Exception {
         final HtmlPage page = webClient.getPage(baseURL + "mvc/jython?name=Jython");
-        assertTrue(page.getTextContent().contains("Hello Jython"));
+        assertTrue(page.getBody().getTextContent().contains("Hello Jython"));
     }
 }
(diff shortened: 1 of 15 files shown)


erdlet:
I'm thinking about additionally implementing some small wrapper around `HtmlPage` to encapsulate the most common operations done on it (e.g. finding the single `h1` by tag name on a page). What are your thoughts on that?

testsuite/src/test/java/org/eclipse/krazo/test/ext/Jsr223IT.java
```diff
@@ -69,8 +79,8 @@ public void testNashorn() {
     // in 2020 (EOL Python 2) we probably need to remove this test with Jython 2.7 anyway
     // and the above mentioned configuration effort would count for nothing.
     @Category(IgnoreOnWildfly.class)
-    public void testJython() {
-        webDriver.navigate().to(baseURL + "mvc/jython?name=Jython");
-        assertTrue(webDriver.getPageSource().contains("Hello Jython"));
+    public void testJython() throws Exception {
+        final HtmlPage page = webClient.getPage(baseURL + "mvc/jython?name=Jython");
+        assertTrue(page.getTextContent().contains("Hello Jython"));
```
chkal:
```java
java.lang.NullPointerException: null
	at org.eclipse.krazo.test.ext.Jsr223IT.testJython(Jsr223IT.java:84)
```

chkal: Looks like the corresponding controller doesn't generate any HTML tags, just plain text content. Maybe that's the cause?

erdlet:
Ah my fault, I just fixed that for Nashorn because I tested only on
Wildfly. I fix this later.

I simply need to check the body for its content

Christian Kaltepoth <notifications@github.com> schrieb am Fr., 10. Jan.
2020, 18:10:

PR #138, opened 2020-01-10
115 update testsuite dependencies

erdlet:


testsuite/src/main/java/org/eclipse/krazo/test/mapper/MvcViewExceptionMapper.java
```diff
@@ -36,6 +37,7 @@ public Response toResponse(MvcViewException exception) {
         Viewable viewable = new Viewable("mvc-error.jsp", models);
 
         return Response.status(492)
+            .type(MediaType.TEXT_HTML_TYPE)
```
chkal: Any idea why this is required? I'm just wondering why the `ViewableWriter` doesn't add this header.

erdlet: Unfortunately not. I receive `application/signed-exchange` as content type and have no idea why.

> I'm just wondering why the ViewableWriter

I assume you think of the `ViewResponseFilter` for setting the headers? The `ViewableWriter` isn't setting any of them. The filter itself isn't hit when running this test.

Anyway, it seems that this is some strange behavior of htmlunit itself, as it worked with the older version and the responded content type is unknown to me. Maybe I have time to look into their code on monday and find a explanation.

chkal: > Unfortunately not. I receive `application/signed-exchange` as content type and have no idea why.

Ok, that's weird. 

> I assume you think of the `ViewResponseFilter` for setting the headers

Not 100% sure where we write this header. I was just wondering why it doesn't seem to work in this case.

Anyway. I don't think there is a real need to investigate this for now. So let's merge this now.

PR #141, opened 2020-01-15
131 annotation support for converter registry

erdlet:


erdlet:
Seems the patched Glassfish makes some trouble.

`org.jboss.arquillian.container.spi.client.container.LifecycleException: Could not connect to DAS on: http://localhost:4848 | Unexpected end of file from server`

I don't have any idea what this could be, so is there someone who saw this problem before?

chkal:
> I don't have any idea what this could be, so is there someone who saw this problem before?

I saw something similar if Glassfish didn't startup correctly. I'll restart the job. Let's see if this fixes the problem.

chkal:
The build is back to green. Seems like this was a temporary issue. I'll review the PR ASAP.

erdlet:
Thanks for your review!

Yeah good idea, I'll add this parameter on monday.

Christian Kaltepoth <notifications@github.com> schrieb am Sa., 18. Jan.
2020, 09:56:

chkal:
> Yeah good idea, I'll add this parameter on monday.

Awesome! Do you want to add this to the current PR or should we merge this one and you create a new one. I'm fine with both options. :-)

erdlet:
I'd add this to the current PR. Shouldn't be much effort to add and test
it, so an additional PR is much bureaucracy for me :)

Christian Kaltepoth <notifications@github.com> schrieb am Sa., 18. Jan.
2020, 10:11:

chkal:
Looks great! Thanks a lot! I'll wait a day or two with merging. Maybe we get more feedback!?

erdlet:
Would be great to get some more feedback :+1:

erdlet:
Thanks a lot!

> This would be an ideal addition to our upcoming user guide (wink wink).

Great idea. I hope to find some time the next days to add documentation for the hidden method filter and this one :)

Issue #115, opened 2019-11-19
Update old dependencies


erdlet:
While checking one of our `pom.xml`s, I recognized a few (really) old dependencies we should update. Fortunately, most of them are test scoped or in our examples and won't affect the productive code.

| Module | Dependency | Current Version (Year) | New Version | Comment |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| core | easymock | 3.3.1 (2015) | 4.1 | [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21350) (:heavy_check_mark:) done in #149 |
| core | weld-se-core | 2.4.8Final (2018) | 3.1.3.Final | [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21351) (:heavy_check_mark:) removed in #149 |
| core | hibernate-validator-cdi | 5.4.1.Final (2017) | 6.1.0.Final | [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21352) (:heavy_check_mark:) done in #149 |
| core | javax.el | 3.0.1-b08 (2015) | 3.0.1-b11 | I have no idea where I found version 4.1, so just upgrade to bugfix release 11 |
| ~examples~ | ~javaee-web-api~ | ~7.0 (2013)~ | ~8.0.1~ | done in  #116  |
| ~examples~ | ~htmlunit~ | ~2.15 (2015)~ | ~2.36~ | [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21353)(:heavy_check_mark:) / done in #133 |
| ~ext~ | ~freemarker~ | ~2.3.28 (2018)~ | ~2.3.29~ | done in #117  |
| ~ext~ | ~handlebars~ | ~4.1.0 (2018)~ | ~4.1.2~ | done in #117 |
| ~ext~ | ~mustache~ | ~0.9.5 (2017)~ | ~0.9.6~ | done in #117 |
| ~ext~ | ~pebble~ | ~3.0.10 (2019)~ | ~3.1.1~ | done in #120 |
| ~ext~ | ~ST4~ | ~4.0.8 (2014)~ | ~4.2~ | [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21354)(:heavy_check_mark:) / done in #137  |
| ~ext~ | ~Thymeleaf~ | ~3.0.10 (2018)~ | ~3.0.11~ | done in #117 |
[...]

PR #149, opened 2020-02-03
Update test dependencies in core module

erdlet:
see: #115

The test dependencies within the core module are up-to-date now. Unfortunately,
Weld changed its behavior and wanted to load servlet related beans which
aren't necessary in this test. As a result, it didn't work anymore. As I don't see
any reason to use CDI for testing the annotation hierachy, I removed CDI and the test
passes again.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -170,12 +170,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.jboss.weld.se</groupId>
-            <artifactId>weld-se-core</artifactId>
-            <version>3.1.3.Final</version>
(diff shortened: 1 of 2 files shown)


chkal:
You could try to rebase and force push top force another ECA test...

erdlet:
> That's why we had to use CDI SE for the tests. However, we change this behavior and therefore getting rid of CDI in the tests is fine for me.

I'll have a look on what happens when I remove the Weld dependency. Maybe we don't need it anymore, that would be fine for me too.

chkal:
Sorry for the delay! Looks great! Merged! :+1:

PR #154, opened 2020-02-11
Update forgotten javax-el implementation

erdlet:
Unfortunately I missed this dependency while updating the core. So here is the last PR for this long list of updates :)

see: #115


chkal:
> Is this a release of the corresponding EE4J project? in this case we don't need a CQ.

This is just a patch release. So my question doesn't make much sense! :-)

Let's merge this one!

Issue #155, opened 2020-02-14
Add testsuite runner for OpenLiberty in Travis


erdlet:
As it seems that IBM Websphere- / OpenLiberty behaves different than TomEE in some cases, we should add a Travis task to run our testsuite against Liberty.

chkal:
Fixed via #156

PR #156, opened 2020-02-14
Add testsuite task for Websphere Liberty

erdlet:
see: #155

To ensure Krazo works on Liberty as well as on TomEE in any case, a new task for Travis is added.
The reason for this is, that Liberty behaves different e.g. when setting classes in the JAX-RS
application which leads to a lot of bug reports and questions.


tck/pom.xml
```diff
@@ -259,7 +259,7 @@
                 <dependency>
                     <groupId>io.openliberty.arquillian</groupId>
                     <artifactId>arquillian-liberty-managed</artifactId>
-                    <version>1.0.3</version>
+                    <version>1.0.6</version>
```
chkal: Do we need a CQ for this?

.travis/tests.sh
```diff
@@ -113,10 +112,7 @@ elif [ "${1}" == "tck-tomee" ]; then
 
 elif [ "${1}" == "tck-liberty" ]; then
 
-  curl -L -s -o wlp.zip "${LIBERTY_URL}"
-  unzip wlp.zip
-  cp .travis/wlp-server-template.xml wlp/templates/servers/defaultServer/server.xml
-  LIBERTY_HOME="$( cd ./wlp/ && pwd )"
+  source .travis/install-liberty.sh
```
chkal: Just a side note: At some point it would be great if we could improve and unify all the scripts that are used to start containers and run the various test suites. Not sure yet how to do that. I just wanted to note, that `tests.sh` is getting huge and difficult to maintain.

erdlet:
I'm not sure if we need a CQ here, as we have this dependency already in
the `tck` module and I just changed the bugfix version.

And I agree that we should clean up and unify as far as possible how we run
our tests etc.

Christian Kaltepoth <notifications@github.com> schrieb am Sa., 15. Feb.
2020, 09:12:

chkal:
> I'm not sure if we need a CQ here, as we have this dependency already in the `tck` module and I just changed the bugfix version.

Ah, ok. I just checked our [wiki page](https://github.com/eclipse-ee4j/krazo/wiki/Creating-CQs) about CQs and it looks like we don't need CQs for patch releases. So let's merge this now! :+1:

Issue #151, opened 2020-02-04
OpenLiberty: CsrfSessionStrategy not working because of missing Session


erdlet:
## Problem description
While debugging issue #147 I recognized that the `SessionCsrfTokenStrategy` is not able to get a `Session` from OpenLiberty where it can extract its CSRF token. This results in a HTTP status 403 and an empty page. Unfortunately, OpenLiberty does not print the stacktrace to the console.

## Possible solution
I assume that this problem can be solved with unwrapping the `HttpRequest` like it is done in `ViewableWriter`.

## Base information
Example Code: https://github.com/hantsy/jakartaee-mvc-sample
Java version: OpenJDK 8

## How to reproduce
1. Start server with `mvn clean package liberty:create dependency:copy liberty:run -Popenliberty`
2. Open http://localhost:9080/jakartaee-mvc-sample/mvc/tasks/new in your browser
3. Submit the form

Issue #157, opened 2020-02-14
OpenLiberty: CXF should call Krazo initialization on startup


erdlet:
While debugging #147 we recognized, that it seems that the Krazo `CoreFeature` is not loaded when configurations are applied in JAX-RS `Application#getClasses`. This problem exists also with Jersey and RESTEasy where it is solved by some custom initalizers like [KrazoJerseyFeature](https://github.com/eclipse-ee4j/krazo/blob/236b5bbed18a43beed1b71ce76ecc1c183608ef4/jersey/src/main/java/org/eclipse/krazo/jersey/bootstrap/KrazoJerseyFeature.java#L40).

chkal:
Just a quick summary of what I think the root cause is:

Krazo uses JAX-RS providers to implement the core functionality and the initialization procedure. This works fine in most cases by annotating classes with `@Provider`. However, if the user implements `Application.getClasses()`, JAX-RS will disable the automatic discovery of providers and Krazo will therefore not initialize correctly.

Jersey and RESTEasy have workaround for this (see [this comment](https://github.com/eclipse-ee4j/krazo/issues/147#issuecomment-584494354)). Maybe we find some way to fix this in CXF as well. If not, the user will have to manually include one or more Krazo related classes in `Application.getClasses()` which may be acceptable if we document it correctly.

erdlet:
I just asked at the CXF mailinglist if there is some kind of interface like Jersey's `ForceAutoDiscovery` or any other mechanism we can use.

erdlet:
Just recognized that there is an issue in the CXF Jira, which asks exactly for an interface like Jersey and RESTEasy are providing.

https://issues.apache.org/jira/browse/CXF-7611

chkal:
Please note that JAX-RS 3.1 (which will most likely be released shortly after Jakarta EE 9, will fix this problem on the spec level:

https://github.com/eclipse-ee4j/jaxrs-api/blob/be10a259518b25bfd4dddaaaf680d425f944722c/jaxrs-api/src/main/java/jakarta/ws/rs/ext/Extension.java

erdlet:
> So it MAY be ok to just document that using Application.getClasses() will force you to also register the corresponding Krazo providers.

I think that is OK for the moment, as there seems no way to fix this in CXF without the upcoming JAX-RS improvement. I'm afraid that CXF will need some time to implement JAX-RS 3.1, but I guess this will happen before implementing the mentioned JIRA issue.

[...]

Issue #132, opened 2020-01-03
SRVE8115W and SRVE8094W warnings at every request in OpenLiberty


mthmulders:
Not sure if this indicates a real problem, but thought I'd report it anyway.

When I deploy my [sample MVC application](https://github.com/mthmulders/openliberty-mvc) to OpenLiberty and I access it through my web browser, the OpenLiberty logs shows this

    [INFO] [WARNING ] SRVE8115W: WARNING: Cannot set status. Response already committed.
    [INFO] [WARNING ] SRVE8094W: WARNING: Cannot set header. Response already committed.  Stack trace of errant attempt to set header: 
    [INFO]         at com.ibm.ws.webcontainer.srt.SRTServletResponse.setHeader(SRTServletResponse.java:1759)
    [INFO]         at [internal classes].(Unknown Source)
    [INFO]         at org.eclipse.krazo.core.ViewableWriter$DelegatingServletOutputStream.write(ViewableWriter.java:248)
    [INFO]         at java.base/java.io.OutputStream.write(OutputStream.java:157)
    [INFO]         at java.base/sun.nio.cs.StreamEncoder.writeBytes(StreamEncoder.java:233)
    [INFO]         at java.base/sun.nio.cs.StreamEncoder.implFlushBuffer(StreamEncoder.java:312)
    [INFO]         at java.base/sun.nio.cs.StreamEncoder.implFlush(StreamEncoder.java:316)
    [INFO]         at java.base/sun.nio.cs.StreamEncoder.flush(StreamEncoder.java:153)
    [INFO]         at java.base/java.io.OutputStreamWriter.flush(OutputStreamWriter.java:254)
    [INFO]         at java.base/java.io.PrintWriter.flush(PrintWriter.java:396)
    [INFO]         at org.eclipse.krazo.core.ViewableWriter.writeTo(ViewableWriter.java:169)
    [INFO]         at org.eclipse.krazo.core.ViewableWriter.writeTo(ViewableWriter.java:70)
    [INFO]         at org.apache.cxf.jaxrs.utils.JAXRSUtils$3.run(JAXRSUtils.java:1479)
    [INFO]         at [internal classes].(Unknown Source)

From what I've seen debugging, the following happens:

[...]

PR #158, opened 2020-02-17
Unwrap HttpRequest and HttpResponse in Session-based CSRF strategy

erdlet:
see: #151


core/src/main/java/org/eclipse/krazo/core/HttpCommunicationUnwrapper.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Unwraps a {@link javax.servlet.http.HttpServletRequest} from a custom wrapper into its original class.
+     *
+     * @param obj the object to unwrap
+     * @param type the target type of the object
+     * @param <T> the generic type of the object
+     * @return the unwrapped {@link javax.servlet.http.HttpServletRequest}
+     */
+    <T extends HttpServletRequest> T unwrapRequest(final T obj, final Class<T> type);
```
chkal: I'm wondering if it makes sense to simplify this to:

    HttpServletRequest unwrapRequest(HttpServletRequest request);

Do we need the generics here?

core/src/main/java/org/eclipse/krazo/core/HttpCommunicationUnwrapper.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Defines an SPI which allows the client to unwrap a HTTP request or response
+ * into a standard {@link javax.servlet.http.HttpServletRequest} or {@link javax.servlet.http.HttpServletResponse}.
+ * <p>
+ * This SPI is intended to be used in case a JAX-RS implementation uses its own wrapper classes for HTTP requests and responses.
+ */
+public interface HttpCommunicationUnwrapper {
```
chkal: Not 100% sure I like the name. But naming is hard. :-)

Maybe something like `ServletUnwrapper`, `ServletClassUnwrapper`? Not sure if this is better though. :-)

cxf/src/main/java/org/eclipse/krazo/cxf/http/CXFHttpCommunicationUnwrapper.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.lang.reflect.InvocationTargetException;
+
+/**
+ * Unwrapper for CXF which extracts the original {@link HttpServletRequest} and / or {@link HttpServletResponse}
+ * from the wrapper which are used e.g. by Liberty. This is required because the wrappers don't use the official wrapper base classes
+ * and therefore Liberty fails to forward such requests because unwrapping isn't possible.
+ */
+public class CXFHttpCommunicationUnwrapper implements HttpCommunicationUnwrapper {
```
chkal: Don't we need `@Priority` to ensure correct ordering?

core/src/main/java/org/eclipse/krazo/core/HttpCommunicationUnwrapper.java
erdlet: I'm not sure, maybe we can do this explicit. I'll check this

core/src/main/java/org/eclipse/krazo/core/HttpCommunicationUnwrapper.java
erdlet: Hmmm I'm struggling with this naming too, but I think `Servlet...` indicates, that we work with servlets and not HTTP requests and responses :thinking:

cxf/src/main/java/org/eclipse/krazo/cxf/http/CXFHttpCommunicationUnwrapper.java
erdlet: Yes we do, I just forgot it (was a long day yesterday :) )

erdlet:
Hmmm, this seems to fail in Liberty for some reason. I'll have a look on my machine what causes this

chkal:
> Hmmm, this seems to fail in Liberty for some reason. I'll have a look on my machine what causes this

But Travis is green! Or am I missing something?

erdlet:
The number of failed tests increased within this PR because of some server error. I need to check this on my local machine to be able to see the server logs

erdlet:
Seems that most of the test failures are caused by #157. I opened #161 to run all testsuite tests with resource autodiscovery, so we know if this change is causing new trouble or not.

erdlet:
I tested the problem from #151 again with the fixed version and the failure didn't occur again.

Anyway, I recognized that the test for the CSRF HTTP Header fails in the testsuite for Liberty, because the header is missing. But when running e.g. the test application from #151, everything works and the header is present in the HTTP response when loading the page. Any idea, what this could be?

core/src/main/java/org/eclipse/krazo/core/HttpCommunicationUnwrapper.java
erdlet: I removed the generics here, works fine :+1:

core/src/main/java/org/eclipse/krazo/core/HttpCommunicationUnwrapper.java
chkal: I agree. While the MVC spec doesn't directly depend on Servlet, Krazo only works in Servlet environments at the moment. Therefore, using Servlet in the name may be ok. However, I'm not strong on this one. So let's keep it as it is.

chkal:
> Anyway, I recognized that the test for the CSRF HTTP Header fails in the testsuite for Liberty, because the header is missing. But when running e.g. the test application from #151, everything works and the header is present in the HTTP response when loading the page. Any idea, what this could be?

Ok, that's really weird. Does this only happen if we unwrap the request in `SessionCsrfTokenStrategy`?

Regarding the other changes proposed in the PR: Everything looks fine. Thanks for applying the modification. The only thing missing: I noticed that a few of the newly created files contain Oracle copyright notices. I guess this was copy & pasted from other files. It would be great if you could adjust the headers.

erdlet:
> Ok, that's really weird. Does this only happen if we unwrap the request in `SessionCsrfTokenStrategy`?

No, this was before this change too. I've no clue what the problem is at this point, as it seems to be something inside the server. I'll try to debug it again, but I'm afraid this is no Krazo specific problem.

> Regarding the other changes proposed in the PR: Everything looks fine. Thanks for applying the modification. The only thing missing: I noticed that a few of the newly created files contain Oracle copyright notices. I guess this was copy & pasted from other files. It would be great if you could adjust the headers.

Thanks for checking the headers, I hope I fixed all of them.

erdlet:
I tested this missing header against JSP and Freemarker and recognized, that this only occurs with JSPs. Maybe this is somehow related to the problem in #132?

PR #161, opened 2020-02-26
Let tests autodiscover their resources

erdlet:
To be able to run the tests on any server without the problem of a
not initialized Krazo feature, they are all set to autodiscover their
resources now. This way, they behave like the TCK tests and we can
test specific behavior (like we have in Jersey or RESTEasy) in own tests


erdlet:
> Should we also do the examples?

+1 for this. I'd propose to do this as part of #28 and finally finish this issue, as it is open for a long time now.

PR #176, opened 2020-05-10
Update handlebars to fix a security issue

gtudan:
... we found in #74. There was nothing we could do back then, so we had to whitelist it.

Upstream fix: https://github.com/jknack/handlebars.java/issues/703

### DIFF
--- a/dependency-check-suppressions.xml
+++ b/dependency-check-suppressions.xml
@@ -26,17 +26,4 @@
         <cve>CVE-2015-9251</cve>
         <cve>CVE-2019-11358</cve>
     </suppress>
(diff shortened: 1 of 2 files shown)


chkal:
Is there already a CQ for this?

gtudan:
Sorry for the delay - it took me a while to figure out if anything but dependencies changed in handlebars.java. It looks like a patch release with lots dependency bumping, but I filed the CQ anyway: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22204

gtudan:
CQ has been approved

PR #177, opened 2020-05-11
Pebble extension maintenance PR

vsmid:
Pebble dependency update, better README.md and Javadoc and better instance creation(for JDKs > 8 removed usage of deprecated Class.forName(...).newInstance() method).

### DIFF
--- a/ext/pebble/README.md
+++ b/ext/pebble/README.md
@@ -2,24 +2,49 @@
 
 An Eclipse Krazo extension for the [Pebble Template Engine][pebble].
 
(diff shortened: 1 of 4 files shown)


erdlet:
Thanks a lot! :) I'll leave this PR open until tomorrow, so others can have a look into it too.

vsmid:
yw :-) Just an info, I noticed some test failures while running on JDK 11.

chkal:
Just a side note. This dependency is a patch-only update. So we won't need a CQ for it.

PR #187, opened 2020-06-12
Handlebars default TemplateLoader was not able to find partials (Fixes #186)

### LINKED ISSUE #186: Handlebars can't load partials
The default TemplateLoader is not able to find/process partials using the `{{> partial}}` format.

jcputney:
Issue: #186

### DIFF
--- a/ext/handlebars/src/main/java/org/eclipse/krazo/ext/handlebars/DefaultHandlebarsProducer.java
+++ b/ext/handlebars/src/main/java/org/eclipse/krazo/ext/handlebars/DefaultHandlebarsProducer.java
@@ -19,6 +19,9 @@
 package org.eclipse.krazo.ext.handlebars;
 
 import com.github.jknack.handlebars.Handlebars;
+import com.github.jknack.handlebars.io.ServletContextTemplateLoader;
+import javax.mvc.engine.ViewEngine;
+import javax.servlet.ServletContext;
 import org.eclipse.krazo.engine.ViewEngineConfig;
 
 import javax.enterprise.inject.Produces;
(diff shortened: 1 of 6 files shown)


erdlet:
Hi @jcputney ,

thanks for your contribution! Please sign the Eclipse Contribution Agreement and sign your commit, to get pipeline green. I'll have a look on your changes in the meantime ;)

ext/handlebars/src/main/java/org/eclipse/krazo/ext/handlebars/DefaultHandlebarsProducer.java
```diff
@@ -29,11 +32,18 @@
...
+    private ServletContext servletContext;
 
     @Produces
     @ViewEngineConfig
     public Handlebars getHandlebars() {
-        return new Handlebars();
+        Handlebars handlebars = new Handlebars();
+        
+        // setup TemplateLoader to allow for finding partials
+        handlebars.with(new ServletContextTemplateLoader(servletContext, ViewEngine.DEFAULT_VIEW_FOLDER));
```
erdlet, line 44: Can you please use the `ViewEngine#VIEW_FOLDER` here, so a possibly customized Folder is used? You find an example on how to load the property in `ViewEngineBase` line 48

erdlet:
> @erdlet I've gone ahead and signed, just waiting on the CI run

Thanks a lot! I just found a small issue regarding the used view folder, but that's no problem to adapt :)

ext/handlebars/src/main/java/org/eclipse/krazo/ext/handlebars/DefaultHandlebarsProducer.java
jcputney: @erdlet let me know if my changes make sense. I added a basic partial test to the existing Handlebars template test. One comment I would add is that it would be nice to get some documentation around how to run the individual tests. I could probably sit down and figure out what profiles to run, but it would be helpful to have a quick example in the README.

erdlet: Thanks for that feedback! You're completely right, this needs to be documented. I hope to find some time for that in the next few days. Anyway, if you want, you can open an issue for that, so it isn't forgotten in case I have no time.

erdlet: The code seems senseful to me, thanks for the fix of the view folder.

PR #193, opened 2020-07-08
Upgrade to Jakarta MVC 1.1-SNAPSHOT

ivargrimstad:
Signed-off-by: Ivar Grimstad <ivar.grimstad@eclipse-foundation.org>

Merging this will require Krazo to be built using `-Pstaging`. 
Not sure if we should set that as a default profile, or maybe it is enough to add instructions to the README.
Travis job will most likely fail as well, any thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,22 @@ services:
 cache:
   directories:
   - $HOME/.m2
+ 
+env:
+     global:
+         - BUILD_PROFILE=${BUILD_PROFILE}
 
 install: /bin/true
 
 jobs:
   include:
     - stage: build
-      script: mvn -B -V -DskipTests -DperformRelease -Dgpg.skip clean install -Psnapshots
+      script: mvn -B -V -DskipTests -DperformRelease -Dgpg.skip clean install ${BUILD_PROFILE}
     - stage: build
-      script: mvn -B -V -DskipTests clean install -Psnapshots
+      script: mvn -B -V -DskipTests clean install ${BUILD_PROFILE}
       jdk: openjdk11
     - stage: test
-      script: mvn -B -V test -Psnapshots
+      script: mvn -B -V test ${BUILD_PROFILE}
     - stage: test
       env: TYPE=glassfish-module
       script: .travis/tests.sh ${TYPE}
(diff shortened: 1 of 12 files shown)


erdlet:
Thanks for the upgrade. The `staging` profile is resolved from the new parent pom, correct?

What would be the benefits of setting the `staging` profile as a default?

ivargrimstad:
Actually, just discovered that we need to activate the `snapshots` profile in order to build against a snapshot of mvc.

The profiles are from the new parent pom, and works like this:

snapshots: consume snapshot dependencies
staging: consume staged artifacts (e.g. release candidates) from the station repository

I suggest we set the snapshots profile default now until we stage a RC of the spec. then we move to that until we release a final version.

erdlet:
> The profiles are from the new parent pom, and works like this:
>
>snapshots: consume snapshot dependencies
>staging: consume staged artifacts (e.g. release candidates) from the station repository
>
>I suggest we set the snapshots profile default now until we stage a RC of the spec. then we move to that until we release a final version.

Thanks for that detailed explanation. Sounds like a good idea. Do you want to adapt the build script in this issue too, so we don't break the master branch build after the upgrade?

ivargrimstad:
I'll give it a try :)

.travis.yml
```diff
@@ -9,15 +9,17 @@ cache:
...
     - stage: build
-      script: mvn -B -V -DskipTests -DperformRelease -Dgpg.skip clean install
+      script: mvn -B -V -DskipTests -DperformRelease -Dgpg.skip clean install -Psnapshots
     - stage: build
-      script: mvn -B -V -DskipTests clean install
+      script: mvn -B -V -DskipTests clean install -Psnapshots
       jdk: openjdk11
     - stage: test
-      script: mvn -B -V test
+      script: mvn -B -V test -Psnapshots
```
erdlet, line 22: Maybe we can store the profile in a variable, as we need to change it when the API gets staged or released later.

.travis/tests.sh
```diff
@@ -9,25 +9,25 @@ if [ "${1}" == "glassfish-bundled" ]; then
 
   curl -L -s -o glassfish5.zip "${GLASSFISH_URL}"
   unzip -q glassfish5.zip
-  mvn -B -V -Pbundled clean install
+  mvn -B -V -Pbundled clean install -Psnapshots
```
erdlet, line 12: Same here, maybe a variable makes sense for easier profile upgrade

.travis.yml
ivargrimstad: I agree, just wanted to see if it worked first. Will add a variable. Thanks!

erdlet: Great, thanks! :+1:

PR #198, opened 2020-07-25
Adding ability to override ScodeID URL param name

jcputney:
Let me know if this implementation works, or if anything needs to be changed.

Signed-off-by: Jonathan Putney <jonathan@putney.io>


chkal:
Hey @jcputney. Thanks a lot for providing this pull request! That's awesome! :+1: 

Some feedback and thoughts below:

(1) I'm wondering if it wouldn't be better if the user could directly configure the full name of the cookie instead of just the prefix used to build it. I know that the current code uses a prefix to build both the request/session attribute name and the name of the cookie. However, customizing the former doesn't make much sense, because it is just used internally. And being able to customize the full cookie name gives the user even more control. So maybe we could simply remove the concept of a "prefix" and add a single constant for the full request/session attribute name + a configuration parameter for the full cookie name. Does this make sense?

(2) We have a class `KrazoConfig` which represents an API for Krazo components to read configuration parameters. I know that this class is not yet used for all configuration parameters, but I would prefer to use it for new ones. So what about creating a method `KrazoConfig.getRedirectScopeCookieName()` which also handles the fallback to a default cookie name if the parameter is not set?

(3) Thanks a lot for also providing a new integration test. This looks great. One minor thing though: The Eclipse Foundation is very strict about copyright headers. And as these are new files, you should replace the two copyright lines with this:

    Copyright (c) 2020 Eclipse Krazo committers and contributors

For existing files you modified, just replace `2019` with `2020` for the `Eclipse Krazo committers and contributors` copyright line.

Thanks a lot for your help with this! :beers:

jcputney:
@chkal any idea why context injection isn't working in RedirectScopeManager? I've rewritten everything to use KrazoConfig, but it's coming back as `null` while running tests.

jcputney:
@chkal ok, I pushed the `KrazoConfig` changes. One thing, when trying to inject `KrazoConfig` into `RedirectScopeManagerInjectionVerifyTest` it was complaining that `KrazoConfig` was not an interface. I ended up extracting `KrazoConfig` into an interface, and renaming the existing class to `KrazoConfigImpl`. Let me know if this works, or if it seems like too much of a shift.

chkal:
Awesome! Thanks a lot! This looks great. Just two comments.

* As mentioned in my previous comment, I don't think that it is worth allowing the developer to configure the name of the request/session attribute. Especially because this is just an internal aspect of the implementation and changing it doesn't have any visible effect. So maybe we should remove `KrazoConfig.getRedirectScopeAttributeName` and replace it with a simple constant in `RedirectScopeManager`. I would prefer not to have a config parameter for this, because developers could be confused otherwise.
* I actually wasn't aware of `RedirectScopeManagerInjectionVerifyTest` and what this test does and how it works. But it uses simple JDK proxies which only work with interfaces. I guess this was the reason you were running into issues. I'm not sure if I like to extract an interface from `KrazoConfig` just to get this very old test to pass correctly. So I guess we could either (a) remove the test as it doesn't bring much benefit at all or (b) replace the JDK Proxy stuff with Mockito/EasyMock which will work with plain classes or (c) keep it as it is in your pull request now. I've a slight preference for (a) or (b). 

@eclipse-ee4j/ee4j-krazo-committers Other thoughts?

@jcputney Thanks for helping us with this!

erdlet:
@jcputney Thanks for that PR! Looks good to me :)

>I actually wasn't aware of RedirectScopeManagerInjectionVerifyTest and what this test does and how it works. But it uses simple JDK proxies which only work with interfaces. I guess this was the reason you were running into issues. I'm not sure if I like to extract an interface from KrazoConfig just to get this very old test to pass correctly. So I guess we could either (a) remove the test as it doesn't bring much benefit at all or (b) replace the JDK Proxy stuff with Mockito/EasyMock which will work with plain classes or (c) keep it as it is in your pull request now. I've a slight preference for (a) or (b).

I'd prefer option b because, as @chkal mentioned, the test doesn't bring much benefit and why should we keep something not really helping us. And maintaining an interface just for one single implementation seems a little bit strange to me (yes, I'm aware of that pattern as good practice, but I think it simply bloats up the code) :)

jcputney:
@chkal Ahh, I guess I misread your first comment. I don't know that I necessarily agree that the Request attribute name is only used internally, because I actually created the original ticket because that name was showing up the URL/address bar upon redirect, and I didn't like the idea of the user being able to see the full name. Not sure if this happens on all systems, but it's happening for me on Wildfly 20.

@erdlet and @chkal, I agree, I wasn't a huge fan of pulling out an interface for such a simple config class. Since the test doesn't seem to be very necessary, I think the cleanest option would just be to remove it.

chkal:
> @chkal Ahh, I guess I misread your first comment. I don't know that I necessarily agree that the Request attribute name is only used internally, because I actually created the original ticket because that name was showing up the URL/address bar upon redirect, and I didn't like the idea of the user being able to see the full name. Not sure if this happens on all systems, but it's happening for me on Wildfly 20.

Thanks for your reply. I was quite sure that the request attribute is only used internally. But I just checked the code, and I was wrong:

https://github.com/eclipse-ee4j/krazo/blob/0b5586fb6e7ee0fe1a7e9d95ae2643d646404da8/core/src/main/java/org/eclipse/krazo/cdi/RedirectScopeManager.java#L265

I'm not super familiar with this class, but it looks like this is some kind of fallback behavior for cases where cookies aren't used. In this case I agree that we should provide a way to customize this as well. So let's keep both configuration parameters. :-)

chkal:
Just another thought. Somehow it is weird that `SCOPE_ID` is used both for the internal request attribute (which is only used internally and has to be unique for the web application) and for the query parameter, which is used only if cookies are not available. Maybe this is a good opportunity to clean this up.

What about this:

* Config Parameter: `org.eclipse.krazo.redirectScopeCookieName`: Used for the name of the cookie
* Config Parameter: `org.eclipse.krazo.redirectScopeQueryParamName`: Used for the name of the query parameter
* An `private static final` constant in `RedirectScopeManager` for the request attribute name.

Thoughts?

jcputney:
@chkal I agree, that makes sense. I'll go ahead and make that change.

jcputney:
Finally got a chance to work on this. Let me know if you see any issues.

chkal:
Unfortunately I'm quite busy today and tomorrow. But I'll definitely find some time to review the PR during the weekend. Sorry for the delay. :disappointed:

PR #205, opened 2020-09-11
Use Eclipse Nexus instead of Sonatype OSSRH for distribution

chkal:
Looks like we are not allowed anymore to push artifacts to Sonatype OSSRH:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=566865

The Eclipse Webmasters suggested trying the Eclipse Nexus instance. As we have to migrate to this one anyway in the long term, let's see if this fixes our current snapshot deployment issue.

Note: The Eclipse Nexus instance is defined in the EE4J parent, so we just have to remove our override.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
     <description>Eclipse Krazo Parent</description>
     <url>https://projects.eclipse.org/projects/ee4j.krazo</url>
 
-     <parent>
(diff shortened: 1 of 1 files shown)


chkal:
I'll merge this directly, so I can verify if the nightly builds are fixed by this.

PR #207, opened 2020-09-30
Activate hidden method filtering by toggle

erdlet:
At the moment, the HiddenMethodFilter is active all the time and checks every
request for possibly changed methods. I could imagine that this is not
wanted by every user, because they maybe want to use other mechanisms to
determine the HTTP method or just use the HTML Form standards. To ensure
that this feature doesn't cause trouble to anyone who isn't aware of the feature,
now a toggle is added to the KrazoProperties and the filter can be activated in the JAX-RS
application.

Signed-off-by: Tobias Erdle <tobias.erdle@innoq.com>

### DIFF
--- a/core/src/main/java/org/eclipse/krazo/KrazoConfig.java
+++ b/core/src/main/java/org/eclipse/krazo/KrazoConfig.java
@@ -102,4 +102,12 @@ public class KrazoConfig {
         return RedirectScopeManager.DEFAULT_QUERY_PARAM_NAME;
     }
 
+    public boolean isHiddenMethodFilterActive() {
+        final Object value = config.getProperty(Properties.HIDDEN_METHOD_FILTER_ACTIVE);
+        if (value instanceof Boolean) {
+            return (boolean) value;
+        }
+
+        return false;
+    }
 }
--- a/core/src/main/java/org/eclipse/krazo/Properties.java
+++ b/core/src/main/java/org/eclipse/krazo/Properties.java
@@ -18,6 +18,8 @@
  */
 package org.eclipse.krazo;
 
+import org.eclipse.krazo.forms.HiddenMethodFilter;
+
 /**
  * Interface Properties. Application-level properties used to configure Krazo.
  *
@@ -54,5 +56,9 @@ public interface Properties {
      * Property for defining default file extension for usage in views
      */
     String DEFAULT_VIEW_FILE_EXTENSION = "org.eclipse.krazo.defaultViewFileExtension";
-
+    
+    /**
+     * Boolean property which enables the {@link HiddenMethodFilter} when set to <code>true</code>.
+     */
+    String HIDDEN_METHOD_FILTER_ACTIVE = "org.eclipse.krazo.hiddenMethodFilterActive";
 }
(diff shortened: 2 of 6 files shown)


chkal:
Looks like the tests broke!? Any idea?

erdlet:
> Looks like the tests broke!? Any idea?

Unfortunately not at the moment. Looks a little bit weird, because the `testsuite` can't resolve the import from `core`. I need to check this today

erdlet:
Puh... running the tests locally with the staging profile doesn't work for me. Arquillian isn't able to resolve the Jakarta MVC API :thinking: Adding the `staging` profile works for the build itself but fails in Arquillian

erdlet:
I got the tests running on my local machine and recognized, that `KrazoConfig` isn't injected into the `HiddenMethodFilter`, even though the filter is configured like others in Krazo. Has anyone an idea what the reason could be?

chkal:
IIRC we had to register **all** CDI beans manually via `KrazoCdiExtension`, because auto-detecting CDI beans didn't work correctly if Krazo is deployed as a Glassfish module.

Also, I'm wondering if we really need `@Provider` on `HiddenMethodFilter`. For all other cases we do NOT use `@Provider` and instead register all providers manually via `DefaultConfigProvider`. IIRC we did so for better supporting JAX-RS applications which implement `Application.getClasses()` which disables JAX-RS's auto-discovery of `@Provider` and other components.

erdlet:
Aaaah thanks. I think this is the missing piece. Before this change I had
no need for injects, so this misconfiguration wasn't recognized. I'll try
it later :)

Christian Kaltepoth <notifications@github.com> schrieb am Fr., 2. Okt.
2020, 18:19:

erdlet:
Nice, thanks @chkal , now the local tests work again (with some adaption for resolving the MVC API) :+1:

erdlet:
I assign this MR to Krazo 2.0.0, as this could be a breaking change to existing applications. I'd propose that we start to write release notes, so nobody is surprised by this change.

erdlet:
I'm afraid we have some weird configuration in our pipeline, as the testsuite wants to use the `krazo-core` from the SNAPSHOT repository :thinking: 

```
Downloading from apache-snapshots: https://repository.apache.org/snapshots/org/eclipse/krazo/krazo-core/1.1.0-SNAPSHOT/maven-metadata.xml
```
This leads to the problem, that the newly created constant isn't present in the tests which leads to the pipeline fail.

chkal:
Good question. Just a guess, but maybe this is caused by the fact that Travis runs each build in a separate VM or docker container.  So the results of the first `build` phase (so what `mvn install` produces) aren't accessible from subsequent build steps (which run in new VMs/containers).

In the testsuite container we just run:

    mvn -P${TYPE} --projects testsuite clean verify ${BUILD_PROFILE}

This runs only the `testsuite` module, but doesn't build `core`, etc. So maybe we have to run something like this before:

    mvn --DskipTests clean install

That's basically like we do it for the TCK as well:

https://github.com/eclipse-ee4j/krazo/blob/1e67a0359bdb7cb4704368c83b040390c39dff50/.travis.yml#L59-L61

and:

https://github.com/eclipse-ee4j/krazo/blob/1e67a0359bdb7cb4704368c83b040390c39dff50/.travis/tests.sh#L86-L103

Just a guess. But this could explain the failures.

However, I'm not entirely happy with the `.travis/tests.sh` script as it grew over time and is difficult to maintain. Cleaning this up has been on my TODO list for quite some time. :disappointed:

erdlet:
Thanks for that hint. I'll try it this evening:) I think that we need to
improve the Pipeline too. It's really complicated now :/

Christian Kaltepoth <notifications@github.com> schrieb am Sa., 3. Okt.
2020, 12:10:

PR #208, opened 2020-09-30
Do some code cleanup

erdlet:
This commit performs some basic code cleanup like removing unused
imports, setting final where possible and improves the usage of the Java
8 Stream API.

Signed-off-by: Tobias Erdle <tobias.erdle@innoq.com>


chkal (approved):
Thanks a lot. Looks great!

TBH: I wasn't aware that we can remove all the `throws IOException` from our filter implementations. I always thought that the `throws` clause must match the interface. Looks like I learned something today. :laughing:

ivargrimstad:
MIssed this one before running the release job, so we'll have to do it after 1.1.0

PR #215, opened 2020-10-10
Fix tests after namespace switch

ivargrimstad:
Not sure why thee BigDecimal conversion test failed after the switch.
Could be Double.NaN is treated differently in Java 8 and 11?
Actually not sure why it worked before either as it looks more correct now :)
Thiis commit is the interesting one: https://github.com/eclipse-ee4j/krazo/pull/215/commits/97dc817f7e114ec0dafe5d30565c879de6e847d5
(The other is just some plugin updates)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ jobs:
   include:
     - stage: build
       script: mvn -B -V clean install ${BUILD_PROFILE}
-#    - stage: test
-#      script: mvn -B -V test ${BUILD_PROFILE}
+    - stage: test
+      script: mvn -B -V test ${BUILD_PROFILE}
 #    - stage: test
 #      env: TYPE=glassfish-module
 #      script: .travis/tests.sh ${TYPE}
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install: /bin/true
 jobs:
   include:
     - stage: build
-      script: mvn -B -V -DskipTests clean install ${BUILD_PROFILE}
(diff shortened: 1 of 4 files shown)


chkal:
I'm pretty sure that `BigDecimal` doesn't support `NaN`, as `NaN` is floating point specific. So that fact that `new BigDecimal("NaN")` is IMO failing is correct. I'm fine with this workaround. I wonder if we run into similar issues in case of `Infinity`, but we can fix this later. :+1:

chkal (approved):

ivargrimstad:
> I'm pretty sure that `BigDecimal` doesn't support `NaN`, as `NaN` is floating point specific. So that fact that `new BigDecimal("NaN")` is IMO failing is correct. I'm fine with this workaround. I wonder if we run into similar issues in case of `Infinity`, but we can fix this later.

Yes, it is weird. But I guess the problem is that it was parsed to Double.NaN, and then the converter didn't recognize NaN as a double, thus blew up. A NaN can be represented with `null` in BigDecimal if we define it that way...

chkal:
> A NaN can be represented with `null` in BigDecimal if we define it that way...

IMO we have two options:

1. represent `NaN` with `null` if bound to a `BigDecimal` and use `NaN` directly if bound to a `Double`.
2. Treat `NaN`, `Infinity`, etc. as invalid for all types and fail conversion (leading to a binding error).

To be honest: I'm not sure which way I prefer. For now (1) would be fine. But I also see good arguments for (2).

Daniel-Dos (approved):

PR #220, opened 2020-11-06
use jcp namespace in xml examples in quickstart

gtudan:
I noticed one of our web.xml examples in the documentation still uses the old "java.sun.com" namespace. We should replace that with the JCP-NS.


Daniel-Dos (approved):

chkal (approved):
Good catch! Thanks a lot!

Issue #221, opened 2020-11-06
Question about metadata-complete in quickstart


gtudan:
I just went through our QuickStart examples and noticed something in the paragraph about metadata-complete: 
https://github.com/eclipse-ee4j/krazo/blob/master/documentation/src/main/asciidoc/_quickstart.adoc#configuration-files

Isn't metadata-complete an attribute of web.xml ? We have it in beans.xml and wildfly is complaining about it:

```
15:13:19,556 WARN  [org.jboss.weld.deployer] (MSC service thread 1-1) WFLYWELD0012: Warning while parsing vfs:/content/partial-service-E-SNAPSHOT.war/WEB-INF/beans.xml:6 cvc-complex-type.3.2.2: Attribute 'metadata-complete' is not allowed to appear in element 'beans'.
```

But I'm not sure about it - might be something specific for one of the other application-servers?

chkal:
> Isn't metadata-complete an attribute of web.xml ? We have it in beans.xml and wildfly is complaining about it:

Very good catch. I think you are correct. The `metadata-complete` should actually go into the `web.xml`.

@erdlet Any thoughts? AFAIK you added this to the documentation?

erdlet:
I can remember this configuration, I think it was part of the discussion in #80. Seems I made a mistake while writing the documentation.

PR #223, opened 2020-11-11
Fix documentation about using metadata-complete

erdlet:
see: #221


gtudan (approved):
Makes sense, thx!

PR #230, opened 2020-11-15
Run testsuite against Wildfly 22

chkal:
This PR reactivates our testsuite and runs it against Wildfly 22 Preview. Looks like this works fine! :-)


chkal:
I'll merge this directly. I guess nobody will complain if we reactivate the testsuite. :laughing:

Issue #250, opened 2021-04-19
RESTEASY003400: Illegal to inject a non-interface type into a singleton after CXF removal


erdlet:
After removing the CXF module and some workarounds, the Jenkins job for Wildfly fails because of the injection of a non-interface type into a singleton. This worked before, so it seems that there was some workaround for CXF which is also necessary for WildFly.

erdlet:
It seems, that the `JaxRsContextProducer` can't be injected by CDI in RESTEasy also and not, as we assumed before, only in CXF.

PR #252, opened 2021-04-19
Load JaxRsContextProducer via BeanManager again

erdlet:
It seems, that the @Context annotation has problems with injecting our
JaxRsContextProducer also on RESTEasy. Unfortunately, the documentation
of @Context itself isn't very helpful, so I just reverted the change and
load the Producer manually again.

see: #250


gtudan (approved):

ivargrimstad (approved):
Jakarta REST 4.0 will deprecate @Context anyway, so it will either be injectable with CDI Inject, or this is the way to go using the manual approach

PR #262, opened 2021-05-16
Fixed dependency scopes and versions

chkal:
I'm working on preparing a release branch for the upcoming 2.0.1 release and found some issues with our dependencies.

  * Most of the Jakarta EE API dependencies in the `core` module are `compile` instead of `provided`-scoped. I guess this was done by mistake during the namespace migration.
  * The `testsuite` and the `ext` modules are still using the Jakarta EE 8 API (`javax.*` version). I guess this didn't cause any problems, because we pull in the individual spec JARs in the correct version as well via some transitive path.
  * The `core` module has a `compile`-scoped dependency on the MVC API. We did this intentionally back then, but I guess it is time to change this and use `provided` scope as for all other APIs.

I ran the TCK for Glassfish and Wildfly locally and everything works fine.

Please note that I'll prepare a separate pull request in the next days to add a `<dependencyManagement>` section to the parent pom to unify the overall dependency handling which is quite confusing at the moment. But I want to get a quick fix in ASAP so we can cherry-pick it into the 2.0.1 release branch.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -117,50 +117,39 @@
             <groupId>jakarta.mvc</groupId>
             <artifactId>jakarta.mvc-api</artifactId>
             <version>${spec.version}</version>
-            <scope>compile</scope>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>jakarta.ws.rs</groupId>
             <artifactId>jakarta.ws.rs-api</artifactId>
             <version>3.0.0</version>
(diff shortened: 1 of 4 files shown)


erdlet (approved):
Thanks! :)

lefloh (approved):

Daniel-Dos (approved):

gtudan (approved):

PR #274, opened 2021-10-18
Jinja2 ViewEngine extension

jelemux:
Hello everyone!

So I played a bit around with Krazo and wanted to try and write my own ViewEngine.
Since there wasn't any Jinja2 ViewEngine yet, I wrote this one.

If you like it, you can merge it. If not, I'll probably put this into its own seperate repository for people to find.
Tell me what you think.

### DIFF
--- a/ext/jinja2/src/main/java/org/eclipse/krazo/ext/jinja2/Jinja2ViewEngine.java
+++ b/ext/jinja2/src/main/java/org/eclipse/krazo/ext/jinja2/Jinja2ViewEngine.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2021 Eclipse Krazo committers and contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.eclipse.krazo.ext.jinja2;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.nio.charset.Charset;
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.hubspot.jinjava.Jinjava;
+import com.hubspot.jinjava.interpret.RenderResult;
+import com.hubspot.jinjava.interpret.TemplateError;
+
+import org.eclipse.krazo.engine.ViewEngineBase;
+import org.eclipse.krazo.engine.ViewEngineConfig;
+
+import jakarta.annotation.Priority;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import jakarta.mvc.engine.ViewEngine;
+import jakarta.mvc.engine.ViewEngineContext;
+import jakarta.mvc.engine.ViewEngineException;
+import jakarta.servlet.ServletContext;
+
+@ApplicationScoped
+@Priority(ViewEngine.PRIORITY_FRAMEWORK)
+public class Jinja2ViewEngine extends ViewEngineBase {
+
+    @Inject
+    @ViewEngineConfig
+    private Jinjava jinjava;
+
+    @Inject
+    private ServletContext servletContext;
+
+    @Override
+    public boolean supports(String view) {
+        return view.endsWith(".j2") || view.endsWith(".jinja") || view.endsWith(".jinja2");
+    }
+
+    @Override
+    public void processView(ViewEngineContext context) throws ViewEngineException {
+
+        Charset charset = resolveCharsetAndSetContentType(context);
+        try (Writer writer = new OutputStreamWriter(context.getOutputStream(), charset);
+             InputStream is = servletContext.getResourceAsStream(resolveView(context));
+             InputStreamReader isr = new InputStreamReader(is, "UTF-8");
+             BufferedReader reader = new BufferedReader(isr)) {
+                 
+            RenderResult result = jinjava.renderForResult(
+                reader.lines().collect(Collectors.joining()), 
+                new HashMap<>(context.getModels().asMap())
+            );
+            if (result.hasErrors()) {
+                throw new ViewEngineException(formatTemplateErrors(result.getErrors()));
+            }
+            
+            writer.write(result.getOutput());
+        } catch (IOException e) {
+            throw new ViewEngineException(e);
+        }
+    }
+
+    private String formatTemplateErrors(List<TemplateError> errors) {
+        StringBuilder builder = new StringBuilder("Errors rendering Jinja2 Template:\n");
+        for (TemplateError error : errors) {
+            builder.append("Severity: ");
+            builder.append(error.getSeverity());
+
+            builder.append("; Line number: ");
+            builder.append(error.getLineno());
+
+            builder.append("; Message: ");
+            builder.append(error.getMessage());
+            builder.append("\n");
+        }
+        return builder.toString();
+    }
+}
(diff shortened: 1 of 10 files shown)


erdlet:
Hi @jelemux,

thanks a lot for the PR. A Jinja2 engine for Krazo is a great addition to the existing ones.

I'll have a deeper look into the code within the next days, hopefully until friday.

erdlet:
@chkal I think this requires a CQ, right?

ivargrimstad:
> @chkal I think this requires a CQ, right?

Correct.
I ran the dash-licenses tool and found the following dependencies that need a CQ:

```
krazo/ext/jinja2$ mvn dependency:list | grep -Poh "\S+:(system|provided|compile)" | sort | uniq -> maven.deps
krazo/ext/jinja2$ java -jar ~/Development/eclipse/dash-licenses/core/target/org.eclipse.dash.licenses-0.0.1-SNAPSHOT.jar maven.deps
[main] INFO Querying Eclipse Foundation for license data for 19 items.
[main] INFO Found 10 items.
[main] INFO Querying ClearlyDefined for license data for 9 items.
[main] INFO Found 9 items.
License information could not be automatically verified for the following content:

maven/mavencentral/com.google.code.findbugs/annotations/3.0.1
maven/mavencentral/com.hubspot.jinjava/jinjava/2.5.2
maven/mavencentral/org.jsoup/jsoup/1.10.3

This content is either not correctly mapped by the system, or requires review.
```

erdlet:
@ivargrimstad thanks for the check.

I'll try to open them this evening.

ext/jinja2/src/main/resources/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
erdlet, line 1: I think you can remove this, as the top level project contains the Apache in its LICENSE file.

ext/pom.xml
```diff
@@ -37,17 +37,18 @@
...
         <module>jade</module>
         <!-- remove until jakarta servlet is supported by jetbrick -->
-<!--        <module>jetbrick</module>-->
+        <!--        <module>jetbrick</module>-->
         <module>jsr223</module>
         <module>mustache</module>
         <module>stringtemplate</module>
         <!-- remove until jakarta is supported by thymeleaf -->
-<!--        <module>thymeleaf</module>-->
+        <!--        <module>thymeleaf</module>-->
```
erdlet, line 48: Can you please remove this new whitespace?

testsuite/src/main/java/org/eclipse/krazo/test/ext/jinja2/HelloController.java
```diff
@@ -0,0 +1,47 @@
...
+import jakarta.mvc.Controller;
+import jakarta.mvc.Models;
+import jakarta.mvc.View;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.QueryParam;
+
+/**
+ * HelloController test.
```
erdlet, line 30: Please remove this unnecessary JavaDoc

testsuite/src/main/java/org/eclipse/krazo/test/ext/jinja2/MyApplication.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.eclipse.krazo.test.ext.jinja2;
+
+import jakarta.ws.rs.ApplicationPath;
+import jakarta.ws.rs.core.Application;
+
+/**
+ * Class MyApplication.
```
erdlet, line 24: Please remove this unnecessary JavaDoc

ext/jinja2/src/main/java/org/eclipse/krazo/ext/jinja2/Jinja2ViewEngine.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Class Jinja2ViewEngine.
+ *
+ * @author Jeremias Weber
+ */
+@ApplicationScoped
+@Priority(ViewEngine.PRIORITY_FRAMEWORK)
+public class Jinja2ViewEngine extends ViewEngineBase {
+
+    protected final Jinjava jinjava = new Jinjava();
```
erdlet, line 54: I'd propose to create `Jinjava` in a separate `Producer` (e.g. like it is done [here](https://github.com/eclipse-ee4j/krazo/blob/master/ext/thymeleaf/src/main/java/org/eclipse/krazo/ext/thymeleaf/DefaultTemplateEngineProducer.java))? 

This way an user can easily overwrite the configuration by using `@Specializes` and doesn't need to extend the whole `ViewEngine`

erdlet (changes requested):

jelemux:
Thanks for the review. I think I'll be able to make the changes tomorrow.

jelemux:
Found a bit of time in-between and made the changes.
Is it ok like that? Would you want me to rebase and squash to create one single commit?

erdlet:
Hi @jelemux,

please update the dependency on `jinjava` to the latest available version. Version `2.5.2` is more than two years old and there are no Git tags for this version in the GitHub repo. Unfortunately, I need the source for exactly this revision for the CQ. Thanks in advance.

Sorry for the delay, I was on vacation and had no real internet connection there.

erdlet (changes requested):
Update on `jinjava` required. Was recognized during CQ process.

jelemux:
I used this version because there was a incompatibility in the dependencies with `guava`. The latest `jinjava` version depends on `guava` version `25.0-jre` and uses the method `com.google.common.collect.ImmutableMap.toImmutableMap` which is not present in the `guava` version imported by `shrinkwrap-resolver-bom` version `3.1.3`. 

Although it seemed the easiest solution at the time, it was not a good idea to just downgrade the `jinjava` version. What I should have done instead (and what I have done now) is upgrade the `shrinkwrap-resolver-bom` version to `3.1.4` (which does not depend on `guava` at all anymore).  
Is this ok for everyone?

erdlet:
Sounds like a plan, thanks for investigating. Please do the update in a separate PR, because this concerns more modules of Krazo

jelemux:
Done. Thanks for the quick response!

erdlet:
Hi @jelemux,
I just merged #275. Can you please rebase this feature branch on the current `master`? I'll create the CQ for Jinjava 2.6.0 and after that we finally should be able to merge this.

erdlet:
CQ created: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23821

erdlet (approved):

erdlet:
CQ seems to be approved. I'll do a last check of the PR this evening and merge it afterwards.

erdlet:
Hi @jelemux,

I ran the testsuite against different application servers (Glassfish and WildFly) and unfortunately it seems that WildFly 25 has problems with Jinjava. Can you please investigate this problem? My log contains the following root cause:

`Caused by: java.lang.ClassNotFoundException: jinjava.jakarta.el.ExpressionFactory`

Thanks!

jelemux:
Hi @erdlet, thanks for pointing that out!
I looked into it but I am honestly at a loss.

First of all, it threw a different exception for me: 
`java.lang.NoClassDefFoundError: Failed to link org/eclipse/krazo/test/ext/jinja2/MyApplication (Module "deployment.5b3a22c8-32e8-4c09-a3e6-5baa2469fb1c.war" from Service Module Loader): jakarta/ws/rs/core/Application`

The only clue I have is that if I remove the `IgnoreOnWildfly` classifier from `AsciiDocIT`, it throws the same exception. 
So maybe it's the same problem as with Asciidoctor? The exception is not very specific though, so could be, could be not...

chkal:
@jelemux First: Thanks a lot for the contribution! :beers: 

> I used this version because there was a incompatibility in the dependencies with `guava`.

I love Guava, but using libraries which depend on different Guava versions is a nightmare. I spent so many hours trying to resolve such issues. :sob: 

> I ran the testsuite against different application servers (Glassfish and WildFly) and unfortunately it seems that WildFly 25 has problems with Jinjava.

@erdlet If the test is green on Glassfish, then maybe it would be fine to use `@IgnoreOnWildfly` for now? We could file a new issue to diagnose this issue later.

erdlet:
Yes, I think we can do this for now. I guess the problem is, as most of the time, namespace switch related. But let's investigate this in an separate issue.

erdlet:
@jelemux can you please just add the `@IgnoreOnWildfly` annotation on this test and open an separate issue? After that I'll merge the PR as it has worked at least on Glassfish on my machine.

chkal:
I finally find some time to work off all the mail notifications I got in the last weeks.

@jelemux Thanks a lot for your contribution. Great work! :beers:

PR #275, opened 2021-11-04
Update shrinkwrap-resolver-bom to latest version

jelemux:
This is needed for #274 to be able to use the latest version of `jinjava`. 

The reason for this is that the latest `jinjava` version depends on `guava` version `25.0-jre` and uses the method `com.google.common.collect.ImmutableMap.toImmutableMap` which is not present in the `guava` version imported by `shrinkwrap-resolver-bom` version `3.1.3`.
`shrinkwrap-resolver-bom` version `3.1.4` does not depend on `guava` anymore, so it should resolve the issue.

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -18,8 +18,8 @@
     SPDX-License-Identifier: Apache-2.0
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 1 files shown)


erdlet:
No CQ required as this is only a bugfix release.

erdlet:
Tested on local machine and seems to work. Thanks for this update!

Issue #277, opened 2021-11-05
Remove Jsr223 extensions as Nashorn engine is removed since JDK 15


erdlet:
The Nashorn engine was deprecated with JDK 11 and removed in JDK 15. I don't think we should provide further support for this. Any thoughts @eclipse-ee4j/ee4j-krazo-committers?

gtudan:
+1 for removing it.

Using it as view engine probably isn’t something that was done very often. 
If someone did and wants to keep using it, it’s just a single class that can easily be copied.

chkal:
+1 for removing it as well.

erdlet:
Thanks for your thoughts. After I got the same feedback on the mailing list I'll remove the extension within the next days.

PR #283, opened 2021-11-10
Remove JSR 223 extension

erdlet:
As the nashorn engine was removed with JDK 15, this engine won't work in
environments using latest technology. As a result of the feedback in the
corresponding issue and on the mailing list, this extension is going to
be removed with this PR.

fixes: #277


chkal (approved):
Sorry for the delay. +1 from my side.

ivargrimstad (approved):

Issue #276, opened 2021-11-05
AvailableConvertersIT fails Glassfish build at Jenkins


erdlet:
Since a while there are errors in the `AvailableConvertersIT` when running Krazo against Glasshfish 6.2.2 on JDK 11 and 17. We need to investigate this.

erdlet:
This behavior is reproducable on my local setup. 

```
java.lang.ClassNotFoundException: javax.servlet.http.HttpServlet
        at org.glassfish.web.loader.WebappClassLoader.loadClass(WebappClassLoader.java:1577)
        at org.glassfish.web.loader.WebappClassLoader.loadClass(WebappClassLoader.java:1427)
        at java.base/java.lang.ClassLoader.defineClass1(Native Method)
        at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
        at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
        at org.glassfish.web.loader.WebappClassLoader.findClass(WebappClassLoader.java:1015)
        at org.glassfish.web.loader.WebappClassLoader.loadClass(WebappClassLoader.java:1544)
        at org.glassfish.web.loader.WebappClassLoader.loadClass(WebappClassLoader.java:1427)
        at org.apache.catalina.core.StandardWrapper.loadServletClass(StandardWrapper.java:1391)
        at org.apache.catalina.core.StandardWrapper.loadServlet(StandardWrapper.java:1257)
        at org.apache.catalina.core.StandardWrapper.allocate(StandardWrapper.java:1069)
        at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:136)
        at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:119)
        at org.apache.catalina.core.StandardPipeline.doInvoke(StandardPipeline.java:611)
        at org.apache.catalina.core.StandardPipeline.invoke(StandardPipeline.java:550)
        at com.sun.enterprise.web.WebPipeline.invoke(WebPipeline.java:75)
        at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:114)
        at org.apache.catalina.connector.CoyoteAdapter.doService(CoyoteAdapter.java:332)
        at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:199)
[...]

PR #284, opened 2021-11-10
Replace test running inside container with remote variant

erdlet:
As it seems that the test was flaky because of the namespace change from
javax -> jakarta and a update of Arquillian and HTMLUnit didn't help, I
decided to replace the test with a similar one which runs outside the
container. This fixes the build errors on Glassfish and gives us the
same information.

fixes: #276

### DIFF
--- a/testsuite/src/main/resources/convert/available/views/index.jsp
+++ b/testsuite/src/main/resources/convert/available/views/index.jsp
@@ -1,0 +1,18 @@
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
+<!doctype html>
+<html>
+<head>
+    <title>Available converter test</title>
+</head>
+<body>
+    <h1>Available converter test</h1>
+
+    <div id="short">${short}</div>
+    <div id="long">${long}</div>
+    <div id="integer">${integer}</div>
+    <div id="float">${float}</div>
+    <div id="double">${double}</div>
+    <div id="biginteger">${biginteger}</div>
+    <div id="bigdecimal">${bigdecimal}</div>
+</body>
+</html>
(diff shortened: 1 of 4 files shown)


testsuite/src/main/resources/convert/available/views/index.jsp
```diff
@@ -0,0 +1,18 @@
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
+<!doctype html>
+<html>
+<head>
+    <title>Available converter test</title>
+</head>
+<body>
+    <h1>Available converter test</h1>
+
+    <div id="short">${short}</div>
```
chkal, line 10: Actually, I'm a bit surprised that this works. Aren't `short`, `double`, etc. reserved keywords? And AFAIK such restrictions apply to EL as well?

chkal (approved):
Looks good to me! Thanks a lot!

testsuite/src/main/resources/convert/available/views/index.jsp
erdlet: Hmm good point. I'll change it just to ensure we won't run into any weird behavior later. Thanks for pointing this out.

chkal: Awesome! Not sure if this is really an issue TBH. It just came up in my head during the review.

ivargrimstad (approved):

ivargrimstad (approved):

testsuite/src/main/resources/convert/available/views/index.jsp
erdlet: Done. Now we shouldn't have any problems with that in the future.

PR #291, opened 2022-01-19
[#233] Reactivating the Thymeleaf extension

### LINKED ISSUE #233: Reactivate the Thymeleaf extension
This extension is currently disabled, because the underlying template engine doesn't support the new `jakarta.*` package namespace yet. We have to:

 * Wait for a new version of the template engine and update our dependencies.
 * Reactivate the corresponding tests in our testsuite (currently disabled via `@Ignore`)

As a last resort we may also decide to remove the extension.

Please note that I assigned this issue to the `2.0.0` milestone. However, the issue should NOT be a blocker for the 2.0.0 release.

jelemux:
This reenables the Thymeleaf extension by using its 3.1.0.M1 milestone release (which uses the jakarta namespace).

To adapt to changes in Thymeleaf, CDIWebContext had to be changed to use the jakarta namespace and the new IWebExchange. Since `JakartaServletWebApplication` has been established to wrap ServletContext in Thymeleaf, there needed to be a producer to inject it into both `DefaultTemplateEngineProducer` and `ThymeleafViewEngine`. But because `JakartaServletWebApplication` is a final class, a wrapper for it had to be written in order to use CDI.

Fixes #233

### DIFF
--- a/ext/pom.xml
+++ b/ext/pom.xml
@@ -2,7 +2,7 @@
 <!--
 
     Copyright (c) 2014-2015 Oracle and/or its affiliates. All rights reserved.
-    Copyright (c) 2018, 2021 Eclipse Krazo committers and contributors
+    Copyright (c) 2018, 2022 Eclipse Krazo committers and contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -43,8 +43,7 @@
 <!--        <module>jetbrick</module>-->
         <module>mustache</module>
         <module>stringtemplate</module>
-        <!-- remove until jakarta is supported by thymeleaf -->
-<!--        <module>thymeleaf</module>-->
+        <module>thymeleaf</module>
         <module>velocity</module>
         <module>pebble</module>
         <module>jinja2</module>
--- a/ext/thymeleaf/src/main/java/org/eclipse/krazo/ext/thymeleaf/CDIWebContext.java
+++ b/ext/thymeleaf/src/main/java/org/eclipse/krazo/ext/thymeleaf/CDIWebContext.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2014-2015 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2018, 2019 Eclipse Krazo committers and contributors
+ * Copyright (c) 2018, 2022 Eclipse Krazo committers and contributors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,16 +20,13 @@ package org.eclipse.krazo.ext.thymeleaf;
 
 import org.thymeleaf.context.IWebContext;
 import org.thymeleaf.context.LazyContextVariable;
+import org.thymeleaf.web.IWebExchange;
 
 import jakarta.enterprise.context.spi.CreationalContext;
 import jakarta.enterprise.inject.Any;
 import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.inject.spi.BeanManager;
 import jakarta.enterprise.util.AnnotationLiteral;
-import jakarta.servlet.ServletContext;
-import jakarta.servlet.http.HttpServletRequest;
-import jakarta.servlet.http.HttpServletResponse;
-import jakarta.servlet.http.HttpSession;
 import java.util.*;
 import java.util.stream.Collectors;
 
@@ -50,9 +47,7 @@ import java.util.stream.Collectors;
  */
 class CDIWebContext implements IWebContext {
 
-    private final HttpServletRequest request;
-    private final HttpServletResponse response;
-    private final ServletContext context;
+    private final IWebExchange exchange;
     private final Locale locale;
 
     private final BeanManager beanManager;
@@ -63,11 +58,9 @@ class CDIWebContext implements IWebContext {
 
     private final Map<String, Object> variables = new HashMap<>();
 
-    CDIWebContext(BeanManager beanManager, HttpServletRequest request, HttpServletResponse response, ServletContext servletContext, Locale locale) {
+    CDIWebContext(BeanManager beanManager, IWebExchange exchange, Locale locale) {
         this.beanManager = beanManager;
-        this.request = request;
-        this.response = response;
-        this.context = servletContext;
+        this.exchange = exchange;
         this.locale = locale;
         this.cdiNamedBeans = enumerateNamedBeans();
     }
@@ -77,26 +70,6 @@ class CDIWebContext implements IWebContext {
         Set<Bean<?>> beans = beanManager.getBeans(Object.class, new AnnotationLiteral<Any>() {
         });
         return beans.stream().map(Bean::getName).filter(Objects::nonNull).collect(Collectors.toSet());
-    }
-
-    @Override
-    public HttpServletRequest getRequest() {
-        return request;
-    }
-
-    @Override
-    public HttpServletResponse getResponse() {
-        return response;
-    }
-
-    @Override
-    public HttpSession getSession() {
-        return request.getSession(false);
-    }
-
-    @Override
-    public ServletContext getServletContext() {
-        return context;
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


chkal:
> I'm currently creating two instances of `JakartaServletWebApplication` (for `DefaultTemplateEngineProducer` and `ThymeleafViewEngine` respectively) which is not pretty, but it should not be the problem here.

Maybe we could simply "produce" an application-scoped instance of `JakartaServletWebApplication` and use it in both classes? Not sure if this works, as I didn't have a detailed look at the code yet.

> The problem is that somehow the request from the `ViewEngineContext` has a `ServletContext` different from the one injected into the ViewEngine and Thymeleaf doesn't like that at all.

It is just a guess, but if we use `@Inject ServletContext`, we will most likely get a CDI proxy class instead of the real `ServletContext` instance. I guess there are a few ways to get a reference to the context:

```
// (A) Directly inject via CDI, but I guess you will get a wrapper
@Inject 
private ServletContext servletContext;

// (B) Inject the request and then call request.getServletContext(). Should return the "real" servlet context?
@Inject 
private HttpServletRequest request;

// (C) Like (B) but Krazo spezific and will return the same request as @Context HttpServletRequest
@Inject 
@JaxRsContext
private HttpServletRequest request;

// (D) via ViewEngineContext.
viewEngineContext.getRequest(HttpServletRequest.class).getServletContext()
```

To make things more complicated: HttpServletRequests can be wrapped via servlet filter. We actually even have some special handling in Krazo to "unwrap" these wrappers. The result is what you will get from `ViewEngineContext`. See:

https://github.com/eclipse-ee4j/krazo/blob/d2467daaa211ae0b8002af54875c68980812a3a0/core/src/main/java/org/eclipse/krazo/core/ViewableWriter.java#L136-L137

Maybe I'll find some time tomorrow to have a look at your branch.

jelemux:
> Maybe we could simply "produce" an application-scoped instance of JakartaServletWebApplication and use it in both classes? Not sure if this works, as I didn't have a detailed look at the code yet.

`JakartaServletWebApplication` is a final class so it cannot be proxied by CDI. 
To solve this, I see two possibilities:
1. Create a wrapper object for `JakartaServletWebApplication` and produce that.
2. Create an issue/PR with Thymeleaf asking them to remove the `final` modifier.
3. Create an issue/PR with Thymeleaf including the `buildExchange()` method (which we need) in `IServletWebApplication`. That way we could just produce and inject that.  
(This would add the need for a wrapper around `HttpServletRequest` and `HttpServletResponse` in Thymeleaf.)

> It is just a guess, but if we use @Inject ServletContext, we will most likely get a CDI proxy class instead of the real ServletContext instance.

Thanks, that seems to be the case. My current (local) codebase uses option D and that works.

jelemux:
I now have it in a working state and pushed the changes. Have a look and tell me what you think.

ext/thymeleaf/src/main/java/org/eclipse/krazo/ext/thymeleaf/JakartaServletWebApplicationWrapper.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+@ApplicationScoped
+public class JakartaServletWebApplicationWrapper {
+
+    private JakartaServletWebApplication application;
+
+    public JakartaServletWebApplicationWrapper() {}
+
+    @Inject
+    public JakartaServletWebApplicationWrapper(final HttpServletRequest request) {
```
jelemux, line 41: Maybe a producer should be used so that people can override it more easily? If so, should it be annotated with `ViewEngineConfig`?

jelemux: @chkal @erdlet should a producer be used and if so, should it be annotated with `ViewEngineConfig`? Maybe I should define an interface for the wrapper so that people can override it?

erdlet: Sorry for the delayed answer. A producer should be enough, as you can overwrite the produced instance with the `@Specialized` annotation.

I hope I'll find some time to have a deeper look into the PR later today.

jelemux:
I don't like my current solution at all. 
I'll try solution 3:

> 3. Create an issue/PR with Thymeleaf including the `buildExchange()` method (which we need) in `IServletWebApplication`.

erdlet:
@chkal after a look into the PR I'm afraid we need a CQ because of the minor update, right? Do you know if we need one for the final 3.1.0 again after creating one for 3.1.0.M1?

ext/thymeleaf/src/main/java/org/eclipse/krazo/ext/thymeleaf/JakartaServletWebApplicationWrapper.java
erdlet: So from my side this looks fine. Only need to check if we have to create a CQ as I asked @chkal below.

erdlet:
> I don't like my current solution at all. I'll try solution 3:
> 
> > 3. Create an issue/PR with Thymeleaf including the `buildExchange()` method (which we need) in `IServletWebApplication`.

Good idea, thanks for doing that 🍻

jelemux:
As we agreed yesterday in the MVC Spec Meeting: Since there is no reaction from Thymeleaf maintainers to my PR, we'll just merge this current solution.

The only open point is the CQ that @erdlet pointed out.
> @chkal after a look into the PR I'm afraid we need a CQ because of the minor update, right? Do you know if we need one for the final 3.1.0 again after creating one for 3.1.0.M1?

ext/pom.xml
```diff
@@ -44,7 +44,7 @@
         <module>mustache</module>
         <module>stringtemplate</module>
         <!-- remove until jakarta is supported by thymeleaf -->
```
chkal, line 46: I guess we can remove this comment now....

ext/thymeleaf/src/main/java/org/eclipse/krazo/ext/thymeleaf/CDIWebContext.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2014-2015 Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2018, 2019 Eclipse Krazo committers and contributors
+ * Copyright (c) 2018, 2019, 2022 Eclipse Krazo committers and contributors
```
chkal, line 3: Just a minor one. AFAIK the Eclipse format for the copyright is `<first_year>, <last_year>`. So there is no need to enumerate all the years. So basically like this:

```suggestion
 * Copyright (c) 2018, 2022 Eclipse Krazo committers and contributors
```

@ivargrimstad Is that correct or do I misunderstand the default convention?

ext/thymeleaf/src/main/java/org/eclipse/krazo/ext/thymeleaf/DefaultJakartaServletWebApplicationWrapperProducer.java
```diff
@@ -0,0 +1,51 @@
...
+import jakarta.servlet.ServletContext;
+import jakarta.servlet.http.HttpServletRequest;
+
+public class DefaultJakartaServletWebApplicationWrapperProducer {
+
+    @Inject
+    private HttpServletRequest request;
+    
+    @Produces
+    @ApplicationScoped
```
chkal, line 34: Maybe we should add the `@Internal` qualifier here? We use this qualifier for some beans used internally by Krazo to protect ourselves against potential conflicts with other libraries which produce the same type as well. We use it for `HttpServletRequest` for example.

I'm not super sure if we should do it here as well. I just wanted to mention it. :-)

chkal (approved):
Hey @jelemux. Sorry for the super long delay. I finally found some time to review your changes.

Overall, it looks great. I just added a few comments, but I would also be fine with merging the current state of the PR. Feel free to have a look at my comments. :-)

ext/thymeleaf/src/main/java/org/eclipse/krazo/ext/thymeleaf/DefaultJakartaServletWebApplicationWrapperProducer.java
jelemux: I tried to understand what the `@Internal` qualifier does but I don't know if I fully get it yet. 
If using it means that the user cannot provide their own producer to override ours, then I would refrain from using it because it might be useful for the user to configure `JakartaServletWebApplication` themselves.

Also if you were afraid that the user might want to inject a `JakartaServletWebApplicationWrapper` somewhere in their code using their own producer: I don't think that this should happen as `JakartaServletWebApplicationWrapper` was written by us specifically for the purpose we are using it for and nothing else.

I hope that clarified things and that I correctly understood what `@Internal` does.

erdlet:
CQ https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23960 created.

erdlet:
CQ is approved, so we can finally merge this PR :)

@jelemux please let me known if you want to squash something before merge or if everything is fine for you.

jelemux:
I already squashed everything, so should be good to go for the merge

PR #296, opened 2022-05-28
Implement specified form method overwrite

erdlet:
This PR implements the newly specified feature 'form method overwrite' and fixes the TCK runner for WildFly on higher JDKs.

As Glassfish ships with a other version of Krazo, the TCK failed there. On WildFly 26 everything went fine. I'd recommend to release a Krazo milestone afterwards which maybe can be put into the Glassfish Beta, so we can test against it.

### DIFF
--- a/core/src/main/java/org/eclipse/krazo/Properties.java
+++ b/core/src/main/java/org/eclipse/krazo/Properties.java
@@ -18,8 +18,6 @@
  */
 package org.eclipse.krazo;
 
-import org.eclipse.krazo.forms.HiddenMethodFilter;
-
 /**
  * Interface Properties. Application-level properties used to configure Krazo.
  *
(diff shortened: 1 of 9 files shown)


erdlet:
I think it makes sense to release a milestone for the TCK first, so I have a version to set in the POM to fix the build.

core/src/main/java/org/eclipse/krazo/Properties.java
```diff
@@ -47,18 +44,12 @@ public interface Properties {
...
 
     /**
      * Property for defining default file extension for usage in views
      */
     String DEFAULT_VIEW_FILE_EXTENSION = "org.eclipse.krazo.defaultViewFileExtension";
-    
-    /**
-     * Boolean property which enables the {@link HiddenMethodFilter} when set to <code>true</code>.
-     */
-    String HIDDEN_METHOD_FILTER_ACTIVE = "org.eclipse.krazo.hiddenMethodFilterActive";
```
chkal, line 63: I'm wondering if we can find a way to preserve the old behavior for at least some time for allowing people easier migration!?! Just thinking out loudly. Or do you think that providing some hints regarding migration would be sufficient?

chkal (approved):
Basically looks fine. Just one comment.

core/src/main/java/org/eclipse/krazo/Properties.java
erdlet: I think providing some hint in the release notes is enough. TBH, I think this feature wasn't used that much, because before the spec change it was an opt-in. So if you used it before, nothing changes.

chkal: Ok. Sure. I'm fine with this as well. We just have to make sure to not forget about the documentation...

chkal:
Update: The CI build currently still fails because of the TCK version.

ivargrimstad:
> I think it makes sense to release a milestone for the TCK first, so I have a version to set in the POM to fix the build.

With release, you mean to staging?
I don't see need to publish a milestone to central.

erdlet:
> With release, you mean to staging?

Yes, I thought in Maven terms ;)

erdlet:
The TCK is now staged in Nexus and the build should run again.

PR #302, opened 2022-06-06
Remove upper version bounds from OSGi Import-Package

chkal:
The same as https://github.com/eclipse-ee4j/mvc-api/pull/79, but for Krazo.


erdlet (approved):
Thanks for that fix!

PR #320, opened 2022-07-22
Jakartaee 10 upgrade

erdlet:
@ivargrimstad I'll check if there are CQs required. If yes, I'll add the tag and link to the CQs.


erdlet:
The dash plugin didn't identify any dependency which needs a CQ.

ivargrimstad:
> The dash plugin didn't identify any dependency which needs a CQ.

Then none is needed. Just go ahead

ivargrimstad (approved):

Issue #317, opened 2022-07-19
Krazo and Jakarta MVC removed from GF7 due to CDI TCK issues


erdlet:
In https://github.com/eclipse-ee4j/glassfish/pull/24029 and https://github.com/eclipse-ee4j/glassfish/pull/24030, Krazo was removed from Glassfish (again) because the CDI TCK won't complete successful otherwise. This is really weird, as MVC only uses the CDI API. We really need to investigate this and find a solution.

erdlet:
I managed to debug the first failing test, which is the `ProcessBeanAttributesNotFiredForBuiltinBean`. This test fails, because Krazo returns a `RequestScoped` version of the `HttpServletRequest` through the `JaxRsContextProducer. HttpServletRequest is [`Default` scoped by default in CDI](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#additional_builtin_beans), so maybe Krazo needs to remove this method? 

@ivargrimstad @chkal what do you think about this? I don't know why we have this producer here, maybe it is superfluous?

erdlet:
I removed the producer and replaced `@Context` with `@Inject` at the places where the `HttpServletRequest` is injected and Krazo works without problems. Bonus: 2 of the 3 classes failing in the CDI TCK are fixed with this.

erdlet:
I could now test what happens in the other failing test (`org.jboss.cdi.tck.tests.full.extensions.lifecycle.atd.AfterTypeDiscoveryTest`). The `KrazoCdiExtension` is adding a bunch of types which are recognized as `alternative` in some cases (see screenshot).
![Uploading Screenshot from 2022-07-25 21-18-06.png…]()

So what we need to is to check why only this few beans are recognized as alternatives and fix it, so the CDI can pass and MVC / Krazo can be added back in GF.

erdlet:
I created issue https://github.com/jakartaee/cdi-tck/issues/400 in the CDI TCK repo. Maybe there we can find some answers what's causing those problems.

erdlet:
The fix in Krazo and the changes made in the CDI TCK lead to a green GF build now! The TCK should be released tomorrow.

erdlet:
Just created PR https://github.com/eclipse-ee4j/glassfish/pull/24052 to re-integrate MVC and Krazo into Glassfish.

erdlet:
We're back on GF again 🥳

PR #321, opened 2022-07-25
Remove producer for HttpServletRequest

erdlet:
It seems that creating a HttpServletRequestBean by Krazo in the JaxRsContextProducer
leads to errors in the CDI TCK because there is already a default for HttpServletRequests
implemented. Removing this producer fixes two CDI tests.

see: #317


erdlet:
Seems like this is / was a spec violation => https://github.com/jakartaee/cdi-tck/issues/400

I'll merge this, so we can proceed to bring Krazo and MVC back into Glassfish.

Issue #319, opened 2022-07-21
Define extensions to keep and discuss extraction into own repository


erdlet:
This issue shall be a base for defining extensions we want to keep maintained by the Krazo team. Also it has to be discussed, if some or all extensions will be moved into a separate repository to reduce maintenance effort in the Krazo repository and to focus on the development of Krazo Core.

To get a quick overview over our extensions and the status of the underlaying template engine, I'll add a list here so we can discuss about them.

@eclipse-ee4j/ee4j-krazo-committers please comment here if there is an extension that must be kept or, on the other hand, can simply be deleted in your opinion

**Edit**: These are all engines supported by an extension with some additional info which may help to decide how to proceed.

| Extension | Template engine | latest version | ext. version | jakarta namespace? | comments | maintainer |
|---------------|------------------------|-------------------|-----------------|----------------------------|----------------|----------------|
| asciidoc | [asciidoctorj](https://mvnrepository.com/artifact/org.asciidoctor/asciidoctorj) | 2.5.4 | 2.5.2 | yes | --- | --- |
| freemarker | [freemarker](https://mvnrepository.com/artifact/org.freemarker/freemarker/2.3.31) | 2.3.31 | 2.3.31 | yes | --- | --- |
| jade | [jade4j](https://mvnrepository.com/artifact/de.neuland-bfi/jade4j/1.3.2) | 1.3.2 | 1.3.2 | yes | looks abandoned and has vulnerabilities | --- |
|mustache | [mustache.java](https://mvnrepository.com/artifact/com.github.spullara.mustache.java/compiler/0.9.10) | 0.9.10 | 0.9.10 | yes | --- | --- |
| pebble | [pebble](https://mvnrepository.com/artifact/io.pebbletemplates/pebble) | 3.1.5 | 3.1.5 | yes | --- | --- |
| stringtemplate | [ST4](https://mvnrepository.com/artifact/org.antlr/ST4) | 4.3.1 | 4.3.3 | yes | --- | --- |
| thymeleaf | [Thymeleaf](https://mvnrepository.com/artifact/org.thymeleaf/thymeleaf) | 3.1.0.M1 | 3.1.0.M2 | yes | --- | @erdlet |
| velocity | [velocity](https://mvnrepository.com/artifact/org.apache.velocity/velocity-engine-core) | 2.3 | 2.3 | yes | --- | --- |
[...]

PR #328, opened 2022-08-05
Remove extensions after their migration to krazo-extensions

erdlet:
see: #319


erdlet:
Silence of reviewers is interpreted as lazy approval. I'll merge this now, because the code is already migrated to [krazo-extensions](https://github.com/eclipse-ee4j/krazo-extensions) and there's no need to keep it here.

Issue #324, opened 2022-07-27
Replace hard-coded priorities with constants from Jakarta Interceptor


erdlet:
We have in a few classes hard-coded priorities which can lead to problems in CDI or misunderstandings in common. To avoid this, we should use constants like `Interceptor.Priority.PLATFORM_AFTER ` and order them by adding some number.

PR #329, opened 2022-08-05
Replace hard coded priorities with Interceptor.Priority

erdlet:
fixes: #324


erdlet:
Merging because reviewer's silence is interpreted as acceptance.

PR #332, opened 2022-08-12
Replace GitHub actions with jenkins

erdlet:


erdlet:
This is the same as I've done in krazo-extensions. Therefore I assume this is OK too and merge it now.

PR #333, opened 2022-08-12
Install compile dependencies in local maven repo

erdlet:
Made a little mistake which only occurs in the testsuite when a core artifact needs a new build. This won't happen with just `mvn compile`, so I made the second step to test and install the artifacts.


erdlet:
Merging directly to fix bug leading to build failures in case of changes to Krazo Core artifacts.

Issue #342, opened 2022-09-05
Fix failing TCK test on WildFly for Krazo 3.0.0


erdlet:
Krazo 3.0.0 fails on WildFly 27 because of this:

```
[ERROR] Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.861 s <<< FAILURE! - in jakarta.mvc.tck.tests.binding.base.BindingBaseTest
[ERROR] submitValidationError(jakarta.mvc.tck.tests.binding.base.BindingBaseTest)  Time elapsed: 0.349 s  <<< ERROR!
java.lang.ClassCastException: class com.gargoylesoftware.htmlunit.xml.XmlPage cannot be cast to class com.gargoylesoftware.htmlunit.html.HtmlPage (com.gargoylesoftware.htmlunit.xml.XmlPage and com.gargoylesoftware.htmlunit.html.HtmlPage are in unnamed module of loader 'app')
	at jakarta.mvc.tck.tests.binding.base.BindingBaseTest.submitForm(BindingBaseTest.java:122)
	at jakarta.mvc.tck.tests.binding.base.BindingBaseTest.submitValidationError(BindingBaseTest.java:90)
```

We need to investigate and fix the issue. The only change happened in the RESTEasy package within the last months was fixing the handling of mixed Jakarta REST Resources in [this commit](https://github.com/eclipse-ee4j/krazo/commit/ce2fa05f0290dc920c26457ee9144d134256bade)

erdlet:
This needs to be backported to a bugfix release of Krazo 3.0.x.

Issue #334, opened 2022-08-15
Krazo disables validation for resources utilizing Content Negotiation


erdlet:
At the moment I try to implement a example application supporting Content Negotiation, so it uses the same Jakarta REST resource for HTML (provided by MVC / Krazo) and JSON content. This works quite well, but when it comes to validation, Krazo disables it for the non-MVC resource because of using this utility method:

```
//org.eclipse.krazo.util.AnnotationUtils

public static <T extends Annotation> boolean hasAnnotationOnClassOrMethod(Class<?> clazz, Class<T> annotationType) {
        return hasAnnotation(clazz, annotationType)
                || Arrays.stream(clazz.getMethods()).anyMatch(m -> hasAnnotation(m, annotationType));
    }
```
This method would work fine if there is a separate controller for MVC and non-MVC resources, but this is exactly not what I want to achieve. So instead, we need to refactor this method to check the invoked method instead.

I did some debugging and saw, that at least the `ValidationInterceptorContext` used in the `org.eclipse.krazo.jersey.validation.KrazoValidationInterceptor` is aware of the request handling method and its annotations via `context.getInvocable().getHandlingMethod().isAnnotationPresent(Controller.class)`, so the refactoring should be easy for this.

As I didn't had a look into the RestEasy implementation yet, I can't say if this is possible there too. I'll create a test application for it too.

erdlet:
I created a reproducer project with RESTEasy and it works there. The `GeneralValidator` contains, in contrast to the Jersey interface, a method called `public boolean isMethodValidatable(Method method)` which seems to decide on the called `Method` directly and handles the request's validation properly. 

Because of that, I'll try to fix the Jersey handling so Jersey and RESTEasy support Content Negotiation in mixed resources.

PR #346, opened 2022-09-30
Fix annotation check for controller methods in RESTEasy

erdlet:
While fixing #334 (PR #335) we introduced a small bug during the check if a method needs to be processed by the general validator or not. Because of this bug, the implementation checked only if a method is annotated with `@Controller`, but not if this is the case for the whole class. The result was Krazo not handling invalid input correctly.

This fix adds the ability to check if the method OR the declaring class is annotated with `@Controller`, so we ensure we really have something that needs to be handled by MVC.

This fix needs to be backported to Krazo 3.0.x.

fixes: #342

### DIFF
--- a/core/src/main/java/org/eclipse/krazo/util/AnnotationUtils.java
+++ b/core/src/main/java/org/eclipse/krazo/util/AnnotationUtils.java
@@ -169,6 +169,20 @@ public final class AnnotationUtils {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


erdlet:
Assuming lazy acknowledgment because of missing response.

chkal (commented):
Good catch! Late +1 for this as well.

PR #358, opened 2022-11-05
Ignore non-MVC requests when validating input

erdlet:
This commit backports the validation issue when using mixed MVC and non-MVC methods. Because the original test is using JDK 11 API, an external HTTP client was added in test scope.

@ivargrimstad  / @chkal please do me a favor and run the tests locally. I think I got weird failures because of my JVM / Glassfish settings.


erdlet:
OK it seems a test is broken in general on the maintenance branch.

erdlet:
Assuming reviewer's silence as acknowledgement. Merging this now.

Issue #364, opened 2022-12-29
Krazo does not work with OpenLiberty 22.* beta (Jakarta EE 10)


mthmulders:
Recently, I was experimenting with Krazo 3.0.1 on Jakarta EE 10, using OpenLiberty 22.0.0.13-beta. I stumbled upon the same error message as reported in #282. I thought it had been resolved with #288, so I checked if the `LibertyHttpCommunicationUnwrapper` got invoked - it did.

So I dug a little further to find the [`ServletUtil#unwrapRequest`](https://github.com/OpenLiberty/open-liberty/blob/366d74a582a310dacd6a51eb3bd8a71109e8c0b6/dev/com.ibm.ws.webcontainer/src/com/ibm/wsspi/webcontainer/util/ServletUtil.java#L64) method of OpenLiberty. As pointed out by @jesse-gallagher in #282, Liberty expects to be able to get an instance of their `IExtendedRequest` type from the passed-in request object. The passed-in request object (supplied by Krazo) is a `HttpServletRequestWrapper` instance backed by a Weld proxy for an [`SRTServletRequest60`](https://github.com/OpenLiberty/open-liberty/blob/366d74a582a310dacd6a51eb3bd8a71109e8c0b6/dev/io.openliberty.webcontainer.servlet.6.0.internal/src/io/openliberty/webcontainer60/srt/SRTServletRequest60.java#L27) instance. The `unwrapRequest` method doesn't handle this object properly, and hence throws the "**SRV.8.2: RequestWrapper objects must extend ServletRequestWrapper or HttpServletRequestWrapper**".

(Aside) the `SRTServletRequest60` does in fact implement the `IExtendedRequest` interface. I think the `isInstance()` check in Liberty fails due to the Weld proxy around it. I've logged an issue for that: https://github.com/OpenLiberty/open-liberty/issues/23774.

[...]

PR #288, opened 2021-12-14
eclipse-ee4j/krazo#282: Added HttpCommunicationWrapper for Liberty JEE9

jesse-gallagher:
This change should fix the trouble observed in https://github.com/eclipse-ee4j/krazo/issues/282 , allowing Krazo to work again on Liberty JEE 9 betas.

The test suites for WildFly and GlassFish should still pass without issue, but there's one test failure in Liberty. I believe it's unrelated to this change, though, and I'll create another issue to identify the trouble there.

PR #365, opened 2022-12-29
Added HttpCommunicationWrapper for Liberty Jakarta EE 10

mthmulders:
Follow-up on #364. This change gets Krazo working on OpenLiberty 22.0.0.13-beta with Jakarta EE 10.

I have added the HttpCommunicationWrapper next to the existing one, introduced in #288. I don't know if they should live together, or if this one should replace the previous one. When living together, I'm not sure about the priority of one vs. the other.

### DIFF
--- a/resteasy/src/main/java/org/eclipse/krazo/resteasy/core/WeldHttpCommunicationUnwrapper.java
+++ b/resteasy/src/main/java/org/eclipse/krazo/resteasy/core/WeldHttpCommunicationUnwrapper.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2022 Eclipse Krazo committers and contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


erdlet:
Thanks for the contribution! I try to run the code against the testsuite within the next days to check if / how we should proceed with the two _unwrapper_ implementations. The code itself looks good👍🏻

chkal:
@mthmulders Thanks a lot for sharing this solution. It would be great if we could improve compatibility with OpenLiberty which such a workaround. Also, it looks to me as if the actual root cause should be fixed on the OpenLiberty side. Maybe we are lucky and OpenLiberty/open-liberty#23774 will be fixed soon.

erdlet:
@mthmulders sorry for the delay. I had to get the testsuite for OL working again. But it seems that this workaround is working great and as soon as the testsuite is merged to master we can merge this PR as well. There will be still test failures, but those are related to some OL internal problems or issues we still have to investigate.

ivargrimstad (approved):

Issue #367, opened 2023-04-30
Add new method to `MvcContext`


erdlet:
https://github.com/jakartaee/mvc/issues/88 defines a new method inside the `MvcContext` that allows easier access to
the name of the hidden method field.

PR #368, opened 2023-04-30
Prepare for 4.0.0 development

erdlet:
Since the new API adds a new interface method to MvcContext, we already need to implement it in this PR.

see: #367

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,8 +47,8 @@
         <spec.version>3.0.0.M1</spec.version>
         <tck.version>3.0.0.M1</tck.version>
 
-        <jersey.version>3.1.0-M8</jersey.version>
-        <arquillian-bom.version>1.7.0.Alpha12</arquillian-bom.version>
+        <jersey.version>3.1.1</jersey.version>
+        <arquillian-bom.version>1.7.0.Alpha14</arquillian-bom.version>
         <shrinkwrap-resolver-bom.version>3.1.4</shrinkwrap-resolver-bom.version>
         <jakartaee-api.version>10.0.0</jakartaee-api.version>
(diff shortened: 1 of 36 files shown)


pom.xml
```diff
@@ -44,8 +44,8 @@
         <maven.compiler.release>11</maven.compiler.release>
 
         <!-- DEPENDENCY PROPERTIES -->
-        <spec.version>2.1.0</spec.version>
-        <tck.version>2.1.0</tck.version>
+        <spec.version>3.0.0.M1</spec.version>
+        <tck.version>3.0.0.M1</tck.version>
 
         <jersey.version>3.1.0-M8</jersey.version>
```
ivargrimstad, line 50: Update jersey version to 3.1.1 ?

ivargrimstad (changes requested):

pom.xml
```diff
@@ -1,4 +1,4 @@
-<!--
+    <!--
```
mthmulders, line 1: Nit: probably not necessary?

pom.xml
erdlet: Yes good catch. Just focused on the spec yesterday. I'll change it

pom.xml
erdlet: Thanks for the review, I'll fix it later👍🏻

chkal (commented):
Other than the two aspects already mentioned, everything looks good to me. Thanks a lot for the effort.

ivargrimstad (approved):

pom.xml
erdlet: Done

ivargrimstad (approved):

chkal (approved):

erdlet:
Removed Krazo and MVC from integration tests to avoid build errors

Issue #369, opened 2023-04-30
Replace `org.eclipse.krazo.defaultViewFileExtension` with API constant


erdlet:
In https://github.com/jakartaee/mvc/issues/38 we standardized the above mentioned constant, so we can replace the Krazo specific one.

##  Tasks

- [ ] The Krazo specific constant is defined in `org.eclipse.krazo.Properties:54` as `DEFAULT_VIEW_FILE_EXTENSION`. It needs to be removed.
- [ ] In `org.eclipse.krazo.KrazoConfig:72` the constant is used inside `getDefaultViewFileExtension()`. Replace it with `jakarta.mvc.engine.ViewEngine#VIEW_EXTENSION`
- [ ] Check if there are any tests covering this issue. If yes: please adapt it. If not: please add one

PR #370, opened 2023-05-10
Replace Krazo default view suffix constant with spec constant

erdlet:
fixes: #369


core/src/main/java/org/eclipse/krazo/core/ViewResponseFilter.java
```diff
@@ -211,7 +211,10 @@ static String appendExtensionIfRequired(String viewName, String defaultExtension
         }
 
         String resultView = viewName;
-        if (!viewName.contains(".")) {
+        if (defaultExtension.contains(".")) {
```
chkal, line 214: Maybe I'm missing something obvious here, but isn't this incorrect? Wouldn't this always add the default extension to the view name, even if the view name already contains a `.`?

I would have expected something like this instead:

```java
if (!viewName.contains(".")) {
  resultView += "." + defaultExtension.replaceAll("^\.", "");
}
```

Or if we want to avoid the regex:

```java
if (!viewName.contains(".")) {
  if(!defaultExtension.startsWith(".")) {
    resultView += ".";
  }
  resultView += defaultExtension);
}
```

erdlet: I'm not sure but I think I wrote test for this and it worked correct🤔 I'll double check it

erdlet: You're right, this would introduce a bug. I added tests and fixed the case when a view already has a view assigned (at least we assume it in case it contains a dot).

ivargrimstad (approved):

chkal:
@erdlet Awesome! Thanks for fixing..

PR #393, opened 2025-12-12
Check for updated MethodValidated class

kalinchan:
In the latest hibernate-validator version, this class has moved resulting in the MVC TCK `ee.jakarta.tck.mvc.tests.binding.base.BindingBaseTest#submitValidationError` test failing in Payara 7 as both interceptors are performing the same validation (Krazo and Hibernate Validator). 

More specifically, the Hibernate Validator interceptor throws a `ConstraintViolationException` preventing the method invocation  to proceed 

cc @ivargrimstad


ivargrimstad:
Thank you!

Issue #398, opened 2026-04-06
Bean validation doesn't work on OpenLiberty


mthmulders:
I've got a [minimal MVC / Krazo application](https://github.com/user-attachments/files/26519200/mvc-playground.zip) set up with a form to enter some information. That information should be validated before being processed.

You can start the app with `mvn liberty:dev` - after that, go to http://localhost:9080/, click "Persons", click "Add". Enter a name and an age below 18. After hitting "Save", the user should see the form again, with some error information, but instead, the user sees a 400 response ("bad request"). In the OpenLiberty log, this message appears

```text
[WARNING ] save.person.age: must be greater than or equal to 18
```

This at least contradicts expectations :-) but more importantly it contradicts @ivargrimstad's answer to my question in #397.

My first impression is that probably, the "standard" bean validation mechanism in Jakarta RESTful Web Services kicks in "too early", preventing Krazo's [`ValidationInterceptor`](https://github.com/eclipse-ee4j/krazo/blob/main/jersey/src/main/java/org/eclipse/krazo/jersey/validation/KrazoValidationInterceptor.java) from being invoked. I had a few breakpoints in that class and they are hit when I do a "GET" retrieve the form (but obviously, all's good then), but they do not get hit when I do the "POST" with invalid data.

PR #416, opened 2026-06-27
Add IT for bean validation

mthmulders:
This PR shows that bean validation for forms works.

* Tested on managed OpenLiberty: failure (as expected from #398).
* Tested on Glassfish: failure 
* Tested on managed Wildfly: failure

It seems to me that the reason for failure is identical on all three servers:

```plain
com.gargoylesoftware.htmlunit.ElementNotFoundException: elementName=[form] attributeName=[name] attributeValue=[form]
	at com.gargoylesoftware.htmlunit.html.HtmlPage.getFormByName(HtmlPage.java:706)
	at org.eclipse.krazo.test.validation.ValidationIT.testFormControllerOk(ValidationIT.java:76)
```

Which might signal that the behaviour as reported in #398 is not specific to OpenLiberty.


ivargrimstad:
Thank you!

A couple of small things.
Please rebase the branch to get the updated Jenkinsfile for the build to pass.
And, it also seems like you have a missing ECA signed associated with the email used. If you have a different email for your Eclipse Foundation accound, please amend the correct one.

mthmulders:
> A couple of small things. Please rebase the branch to get the updated Jenkinsfile for the build to pass.

Done

> And, it also seems like you have a missing ECA signed associated with the email used. If you have a different email for your Eclipse Foundation accound, please amend the correct one.

I noticed on the first run, updated my details with the Eclipse Foundation, but that didn't trigger re-validation. The force-with-lease commit that I just pushed seems to have triggered the re-validation though.

